PR #6, opened 2017-04-02
Basic WSDL generation

KaiWalter:
I created a basic WSDL rendering. Currently only string and int input and output parameters are supported but it should be easy to enhance.


DigDes:
Thanks for contributing! Updated nuget package: [https://www.nuget.org/packages/SoapCore/0.4.0]()

KaiWalter:
My pleasure. We want to use SoapCore here (connecting SAP ABAP stack to Azure Service Fabric) I will continue to enhance it and get back to U guys.

PR #26, opened 2017-11-17
Fix Serializer and Path issue

Anoyomouse:
This PR fixes two issues i found while using your library:
- Fixed XML Serializer not using default namespace forcing the output XML to emit it's namespace
- Fixed path problem when not hosted at /, you need PathBase to give the extra part of the path, so the soap endpoint winds up at the correct place


kotovaleksandr:
Thanks for help!
Version with your changes: [nuget](https://www.nuget.org/packages/SoapCore/0.9.1)

PR #51, opened 2017-12-12
Refactoring for later changes

MFinkBK:
Here's another change, preparing code for another change about adding authorisation. It's just a little refactoring of the exception handling. I changed

```
while (exception.InnerException != null)
	exception = exception.InnerException;
```
to just use
```
var errorText = exception.InnerException != null ? exception.InnerException.Message : exception.Message;
```

Maybe it's even better to just use exception.ToString(), so that every contained exception is logged correctly.


matafonoff:
Using exception.ToString() could be a little bit distructive for PROD environment due to full stack information of error.

PR #72, opened 2018-03-30
Support for void methods

bhenden:
Handles response tasks without any result.


kotovaleksandr:
Please, resolve conflicts and resend PR.

PR #103, opened 2018-07-12
add soap model validator

mdolega:
I think its good to have some kind of interceptor of requests before service method would be run.


kotovaleksandr:
Please add tests and fix exists tests.

PR #111, opened 2018-08-03
fix extra xml response element

DmitryBryluk:
for old style calls with single return value and input param there is extra result tag for XmlSerializer:
```
SomeMethodResponseType SomeMethod(SomeMethodRequestType)

<SomeMethodResponseType>
    <SomeMethodResult> <= extra element
        ... contents of SomeMethodResponseType ...
    </SomeMethodResult> <= extra element
</SomeMethodResponseType>
```

fixed this

### DIFF
--- a/src/SoapCore/ServiceBodyWriter.cs
+++ b/src/SoapCore/ServiceBodyWriter.cs
@@ -29,7 +29,11 @@ namespace SoapCore
 
 		protected override void OnWriteBodyContents(XmlDictionaryWriter writer)
 		{
-			writer.WriteStartElement(_envelopeName, _serviceNamespace);
+			// do not wrap old-style single element response into additional xml element for xml serializer
+			var needResponseEnvelope = _serializer != SoapSerializer.XmlSerializer || _result == null || (_outResults != null && _outResults.Count > 0);
+
+			if (needResponseEnvelope)
+				writer.WriteStartElement(_envelopeName, _serviceNamespace);
 
 			if (_outResults != null)
 			{
@@ -82,7 +86,8 @@ namespace SoapCore
 					case SoapSerializer.XmlSerializer:
 						{
 							// see https://referencesource.microsoft.com/System.Xml/System/Xml/Serialization/XmlSerializer.cs.html#c97688a6c07294d5
-							var serializer = CachedXmlSerializer.GetXmlSerializer(_result.GetType(), _resultName, _serviceNamespace);
+							var resultType = _result.GetType();
+							var serializer = CachedXmlSerializer.GetXmlSerializer(resultType, needResponseEnvelope ? _resultName : resultType.Name, _serviceNamespace);
 							lock (serializer)
 								serializer.Serialize(writer, _result);
 						}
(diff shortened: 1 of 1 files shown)


ramondeallima:
Does it still work?

I'm getting it as extra tags yet.

Do I need to change anything in my code?

ramondeallima:
@DmitryBryluk  

Does it still work?

I'm getting it as extra tags yet.

Do I need to change anything in my code?

DmitryBryluk:
try looking https://github.com/DigDes/SoapCore/blob/master/src/SoapCore.Tests/Serialization/Models.Xml/ISampleService.cs

there are several ways to do what you need, at least one of them have to work

ramondeallima:
Thanks bro! I will verify

ramondeallima:
Case somebody have this problem, the code below worked for me.

`
		// old style call, when all in/out params are packed into single request/response params		// both styles are completely compatible, if we have same method name		//   and req/resp classes as {MethodName}{Request|Response}		// produced/consumed xml response is compatible with legacy wcf/ws		// this operation contract is exactly as in legacy wcf/ws sources and can be consumed there		// to avoid extra envelope element, mark request/reponse type as [MessageContract] and do not use ref/out params		//   otherwise compatibility will be broken		//   (effective only for XmlSerializer)		[OperationContract(Action = ServiceNamespace.Value + nameof(PingComplexModelOldStyle), ReplyAction = "*")]		[XmlSerializerFormat(SupportFaults = true)]		PingComplexModelOldStyleResponse PingComplexModelOldStyle(			PingComplexModelOldStyleRequest request);

`

ramondeallima:
@DmitryBryluk 

Thanks for your last help, I managed to solve it with one of the examples.

image

However, I have a similar problem for the request XML generated in SoapUI.

The tags are:

<! - Optional: ->

<has: model>

I managed to remove the comment <! - Optional: -> using the decorator [DataMember (IsRequired = true)] with DataContractSerializer, but I haven't found anything similar for XmlSerializer.

But the main thing would be to remove the <has:model>.

I tried some examples of the last link, but nothing worked, I believe it's just for the reponse

DmitryBryluk:
perhaps it is just wrong contract generation by SoapCore, if you are talking about what SoapUI auto generated from your endpoint. just fill correct xml disregarding what SoapUI got from SoapCore and it have to be fine. otherwise no idea

wahajahmedofficial:
It is still cant be fixed for me. i am still having that extra result field. it would be great if you can upload an example code so @DmitryBryluk

natghi2010:
> @DmitryBryluk
> 
> Thanks for your last help, I managed to solve it with one of the examples.
> 
> image
> 
> However, I have a similar problem for the request XML generated in SoapUI.
> 
> The tags are:
> 
> <! - Optional: ->
> 
> <has: model>
> 
> I managed to remove the comment <! - Optional: -> using the decorator [DataMember (IsRequired = true)] with DataContractSerializer, but I haven't found anything similar for XmlSerializer.
> 
> But the main thing would be to remove the has:model.
> 
> I tried some examples of the last link, but nothing worked, I believe it's just for the reponse

how did you do it?

PR #112, opened 2018-08-06
fix attributes in xml out params

DmitryBryluk:
purpose - proper serialize out/ref params in response, include attributes, custom param name and its type for XmlSerializer. see outComplexParam example below. ComplexParamType fields are marked as [XmlAttribute] and missing in original implementation.

```
[ServiceContract(Namespace = ServiceNamespace ConfigurationName = "...")]
[XmlSerializerFormat(SupportFaults = true)]
public interface ISomeService
{
    [OperationContract(Action = ServiceNamespace + nameof(SomeMethod), ReplyAction = "*")]
    [XmlSerializerFormat(SupportFaults = true)]
    bool SomeMethod(
        OtherComplexParamType inComplexParam,
        int inSimpleParam,
        out int outSimpleParam,
        out ComplexParamType outComplexParam);
}

<s:Envelope xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:s="http://schemas.xmlsoap.org/soap/envelope/">
   <s:Body>
      <SomeMethodResponse xmlns="http://ournamespace.net/webservices/">
         <SomeMethodResult>true</SomeMethodResult>
         <outSimpleParam>simple_value</outSimpleParam>
         <outComplexParam xsi:type="ComplexParamType" Field1="..." Field2="..." ... />
      </SomeMethodResponse>
   </s:Body>
</s:Envelope>
```

### DIFF
--- a/src/SoapCore/ServiceBodyWriter.cs
+++ b/src/SoapCore/ServiceBodyWriter.cs
@@ -58,13 +58,20 @@ namespace SoapCore
 							switch (_serializer)
 							{
 								case SoapSerializer.XmlSerializer:
-									// todo: write element with outResult.Key name and type information outResultType
+									// write element with name as outResult.Key and type information as outResultType
 									// i.e. <outResult.Key xsi:type="outResultType" ... />
-									// upd: custom element name is ok, missing type info
 									var outResultType = outResult.Value.GetType();
 									var serializer = CachedXmlSerializer.GetXmlSerializer(outResultType, outResult.Key, _serviceNamespace);
 									lock (serializer)
-										serializer.Serialize(writer, outResult.Value);
+										serializer.Serialize(stream, outResult.Value);
+									// add outResultType. ugly, but working
+									stream.Position = 0;
+									XmlDocument xdoc = new XmlDocument();
+									xdoc.Load(stream);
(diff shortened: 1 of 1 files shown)


kotovaleksandr:
Hi Dmitry! Thanks for help! Why need lock for XmlSerializer? XmlSerializer is not thread safe?
And, please add tests for new functionality. Thanks!

DmitryBryluk:
@kotovaleksandr,
> Why need lock for XmlSerializer? XmlSerializer is not thread safe?

seems it is thread-safe, but why we are locking it [here](https://github.com/DigDes/SoapCore/blob/master/src/SoapCore/ServiceBodyWriter.cs#L91)?

>please add tests for new functionality

asap

kotovaleksandr:
> Why need lock for XmlSerializer? XmlSerializer is not thread safe?
> 
> seems it is thread-safe, but why we are locking it here?
I don't know why:)

kotovaleksandr:
Check https://www.nuget.org/packages/SoapCore/0.9.8.4
Please add tests for both PR, thanks!

PR #120, opened 2018-08-14
restore compatibility and some fixes

DmitryBryluk:
done:
- restored compatibility with original edition of SoapCore
- restored original samples, to be simple and readable
- moved all tests to root test folder (still using xunit, but it runs smoothly with other tests)
- moved extended legacy samples into tests
- removed reference to samples from tests
- fixed [MessageContract] request/response handling

some notes:

DataContractSerializer and XmlSerializer attributes markup seems not compatible together, and DataContractSerializer can't handle complex cases. not much time to explore why. but with these tests it will be easier to explore what is happening.

all possible cases (including incompatibility) are shown in tests for both serializers


DmitryBryluk:
@kotovaleksandr will you update nuget?

kotovaleksandr:
@DmitryBryluk https://www.nuget.org/packages/SoapCore/0.9.8.6

PR #129, opened 2018-09-06
Ignoring props for WSDL with [IgnoreDataMember] attribute set

neilsimp1:
Modifed MetaBodyWriter.cs to exlcude any properties with [IgnoreDataMember] on them were omitted from the WSDL.

Example:
```
public class User {
    public long Id { get; set; }
    public string Username { get; set; }
    [IgnoreDataMember]
    public string Name { get; set; } // This property will not appear in the WSDL
    public int Age { get; set; }
}
```

### DIFF
--- a/src/SoapCore/MetaBodyWriter.cs
+++ b/src/SoapCore/MetaBodyWriter.cs
@@ -165,7 +165,7 @@ namespace SoapCore
 					}
 					else
 					{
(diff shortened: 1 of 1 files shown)


neilsimp1:
@kotovaleksandr Can this update be added to NuGet? I need this update and it's most easily installed via NuGet.

PR #154, opened 2018-11-06
Add XElement support for WSDL

gunnarif:
Added support for XElement for WSDL. 
Generates an Any element that allows element from any namespace in the soap request


drabb1956:
I'm trying to include an XElement in a class that I'm returning from a SoapCore server and it is causing a stack overflow exception in an unknown class when I attempt to update the service reference from the consuming application.

PR #158, opened 2018-11-16
Support for [ServiceContract] without Name attribute

MFinkBK:
I'm using SoapCore to implement a Soap WebService that must be compatible with a pre-WCF Soap WebService implementation from long ago. The old web service is marked with [ServiceContract] and specifies a Namespace, but no class name. This results in WebService action URLs like this:

`http://tempuri.org/operationName`

So there's no contract name (the classes name) between namespace and operation name. The following PR fixes generating the proper <soap:operation soapAction=""> tag attribute when no name is present in the ServiceContract C# attribute.


MFinkBK:
Never mind, this is not the right fix for my problem.

PR #159, opened 2018-11-18
Add interface IServiceOperationTuner that helps to some extra tuning …

Avemey:
…for each

operation call.
Method Tune() from that interface takes 3 parameters:
 * httpContext - current http context
 * serviceInstance - instance of service
 * operation - operation description

Use IServiceOperationTuner to get some information from http context
for operation call.
It's helps to migrate from old asmx to .net core SOAP implementation when
not possible to change clients for backward compatibility and clients sends additional
parameters in http headers. It is sad but it is not rare situation in real life.
Using additional midlleware is more complex and not clear solulion.

Added new interface instead of extend of IMessageInspector to avoid
break working solution for SoapCore users.

Extension AddSoapServiceOperationTuner() for IServiceCollection helps to
add operation tuner.

Note: use some thread sinhronization stuff (like ThreadLocal) if service is register as  singleton.

### DIFF
--- a/src/SoapCore.Tests/ITestService.cs
+++ b/src/SoapCore.Tests/ITestService.cs
@@ -60,5 +60,8 @@ namespace SoapCore.Tests
 		[OperationContract]
 		[ServiceFilter(typeof(ActionFilter.TestActionFilter))]
 		ComplexModelInput ComplexParamWithActionFilter(ComplexModelInput test);
+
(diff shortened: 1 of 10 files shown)


kotovaleksandr:
Hi @Avemey thanks for your help! Can you add some description in markdown file with examples?

PR #172, opened 2018-12-06
Datacontract metadata

andradf:
This PR fixes (some) issues with the metadata generation for DataContractSerialization. SvcUtil now successfully creates a proxy for a DataContractSerialization service. It creates a new BodyWriter and is based largely on MetaBodyWriter and the metadata output from a WCF SOAP service.
The main changes relative to MetaBodyWriter are:

- custom objects are placed in a new namespace http://schemas.datacontract.org/2004/07/{ServiceNamespace}
- action inputs must be nillable for svcutil to generate leaner proxy methods (without wrapper request and body classes)
- DataContractSerializer requires sequence ordering to respect some predefined rules, so far I've only implemented one of these rules (alphabetical) which I assume will be the most commonly used one.
- wsam:Action attributes in wsdl:input and wsdl:output attributes
- new type schema http://schemas.microsoft.com/2003/10/Serialization/. This schema is static, so it's just a copy paste from the metadata generated by WCF.
- refactoring of System Array types

There weren't any tests implemented for MetaBodyWriter, so I didn't implement any on mine, but any suggestions on how to test this are welcome. There are still some pending issues. Right now the attributes DataElement and DataContract are not examined, so custom property names are not supported. Sequence ordering for classes with inheritance and properties with Order in the DataMember attribute is not implemented yet. And building a dependency graph between schemas to clean up imports should probably be done too. There's a lot of code repeat between MetaBodyWriter and MetaWCFBodyWriter, I didn't want to touch existing functionality so I left MetaBodyWriter mostly unchanged, but maybe they should both be derived from the same base class that exposes some helper methods?

I also used C# 7 value tuples, but since StyleCop doesn't yet support C#7 on their non-pre-release packages, I had to pragma out some warnings.

### DIFF
--- a/src/SoapCore/MetaBodyWriter.cs
+++ b/src/SoapCore/MetaBodyWriter.cs
@@ -78,6 +78,7 @@ namespace SoapCore
 
 		private void AddTypes(XmlDictionaryWriter writer)
 		{
+			writer.WriteStartElement("wsdl:types");
 			writer.WriteStartElement("xs:schema");
 			writer.WriteAttributeString("xmlns:xs", XMLNS_XS);
 			writer.WriteAttributeString("elementFormDefault", "qualified");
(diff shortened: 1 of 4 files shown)


andres-carrilloAdapt2:
Is there doc on how to use this in our Conracts so that when we generate wsdl it doesnt break the existing clients?

PR #175, opened 2018-12-14
Update Middleware to use trailing path

espcoder:
We are using a url with a dynamic path part which is not known at service startup time. There are too many variations of the path to register them all as service endpoints. This PR changes the middleware Invoke to compare the path to a service endpoint using only the trailing part of the path.

Example:
Url: http://localhost/AppName/DynamicPathPart/Service.asmx
compare on trailing path: "/Service.asmx"  
instead of path: "/DynamicPathPart/Service.asmx"


kotovaleksandr:
@espcoder i understand what problem you solve, but PR break common behavior (for example #187).
Please, add setting or change code to solve your problem and save previous behavior with tests.

Issue #206, opened 2019-03-07
Parent classes not inherited from in WSDL


agpreynolds:
Given the following basic class structure:

```
    public abstract class Animal
    {
        public Guid RequestId { get; set; }
    }

    public class Cat : Animal
    {
        public string Name { get; set; }
    }

    public class Dog : Animal
    {
        public string Name { get; set; }
    }
```

and the following service definition:

```
    [ServiceContract]
    public interface ITestService
    {
        [OperationContract]
        Cat GetCat();

        [OperationContract]
        Dog GetDog();

        [OperationContract]
        bool CheckAnimalStatus(Animal animal);
    }
```

It should be possible for the consumer to do something like the following:

```
service.OpenAsync().Wait();

var cat = service.GetCatAsync().Result;

while (service.CheckAnimalStatusAsync(cat).Result)
{
       Console.WriteLine($"Cat id: {cat.RequestId}, name: {cat.Name} is ok.");
}

service.CloseAsync().Wait();
```

However the current serialization strategy is to serialize all properties, including properties from a parent class into one class, see:

```
<xs:complexType xmlns:ser="http://schemas.microsoft.com/2003/10/Serialization/" name="Cat">
<xs:sequence>
<xs:element minOccurs="0" name="Name" nillable="true" type="xs:string"/>
<xs:element minOccurs="0" name="RequestId" type="ser:guid"/>
</xs:sequence>
</xs:complexType>
<xs:element name="Cat" nillable="true" type="tns:Cat"/>
<xs:complexType xmlns:ser="http://schemas.microsoft.com/2003/10/Serialization/" name="Dog">
<xs:sequence>
<xs:element minOccurs="0" name="Name" nillable="true" type="xs:string"/>
<xs:element minOccurs="0" name="RequestId" type="ser:guid"/>
</xs:sequence>
</xs:complexType>
<xs:element name="Dog" nillable="true" type="tns:Dog"/>
```

This results in a class looking like the following being generated:

```
    [System.Diagnostics.DebuggerStepThroughAttribute()]
    [System.CodeDom.Compiler.GeneratedCodeAttribute("dotnet-svcutil", "1.0.0.1")]
    [System.Runtime.Serialization.DataContractAttribute(Name="Cat", Namespace="http://schemas.datacontract.org/2004/07/CoreService")]
    public partial class Cat : object
    {
        
        private string NameField;
        
        private System.Guid RequestIdField;
        
[...]

PR #209, opened 2019-03-08
Add support for serializing extension types

agpreynolds:
Addresses #206 

Add section to WriteComplexType method to add an xs:complexContent section to complex types with base types.

Tweak the method of property discovery - now only properties defined in a class are serialized to that class, inherited properties will be serialized via inheritance.

This means the following class structure and service definition:

```
[ServiceContract]
    public interface ITestService
    {
        [OperationContract]
        Cat GetCat();

        [OperationContract]
        Dog GetDog();
    }
```

```
public abstract class Animal
    {
        public Guid RequestId { get; set; }
        public Group Group { get; set; }
    }

    public class Cat : Animal
    {
        public string Name { get; set; }
    }

    public class Dog : Animal
    {
        public string Name { get; set; }
        public Breed Breed { get; set; }
    }

    public enum Breed { Terrier, Hound, Puppy }
    public enum Group { Mammal, Amphibian, Reptile }
```

Produces the following WSDL classes:

```
<xs:complexType xmlns:ser="http://schemas.microsoft.com/2003/10/Serialization/" name="Cat">
<xs:complexContent>
<xs:extension base="tns:Animal">
<xs:sequence>
<xs:element minOccurs="0" name="Name" nillable="true" type="xs:string"/>
</xs:sequence>
</xs:extension>
</xs:complexContent>
</xs:complexType>
<xs:element name="Cat" nillable="true" type="tns:Cat"/>
<xs:complexType xmlns:ser="http://schemas.microsoft.com/2003/10/Serialization/" name="Dog">
<xs:complexContent>
<xs:extension base="tns:Animal">
<xs:sequence>
<xs:element nillable="true" type="tns:Breed" name="Breed"/>
<xs:element minOccurs="0" name="Name" nillable="true" type="xs:string"/>
</xs:sequence>
</xs:extension>
</xs:complexContent>
</xs:complexType>
<xs:element name="Dog" nillable="true" type="tns:Dog"/>
```

Only bit i'm not happy with is the method of detecting a root class, type.BaseType.Name != "Object" seems a bit rubbish

### DIFF
--- a/src/SoapCore/MetaWCFBodyWriter.cs
+++ b/src/SoapCore/MetaWCFBodyWriter.cs
@@ -452,7 +452,16 @@ namespace SoapCore
 				return;
 			}
 
-			foreach (var property in type.GetProperties().Where(prop => prop.CustomAttributes.All(attr => attr.AttributeType.Name != "IgnoreDataMemberAttribute") && !prop.PropertyType.IsPrimitive && !SysTypeDic.ContainsKey(prop.PropertyType.FullName)))
+			if (HasBaseType(type))
+			{
+				DiscoveryTypesByProperties(type.BaseType, false);
+				_complexTypeToBuild.Enqueue(type.BaseType);
+			}
+
+			foreach (var property in type.GetProperties().Where(prop =>
+						prop.DeclaringType == type
+						&& prop.CustomAttributes.All(attr => attr.AttributeType.Name != "IgnoreDataMemberAttribute")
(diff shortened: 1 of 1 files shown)


agpreynolds:
Hi, out of interest what type is being written that has a base type of null? I will add the null check and raise another PR

psalmnoelff:
Hi guys, i was trying this feature out but it seems that its not being rendered as complex types in the xml/wsdl. Is there a config where I need to enabled this?

PR #221, opened 2019-03-18
Feature/trailing path

espcoder:
Create an opt-in service tuner for using the last path-part of an endpoint address. This allows dynamic paths to be sent by the client.

### DIFF
--- a/src/SoapCore.Tests/TrailingServicePathTuner/MockMessageEncoder.cs
+++ b/src/SoapCore.Tests/TrailingServicePathTuner/MockMessageEncoder.cs
@@ -1,0 +1,32 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.ServiceModel.Channels;
+using System.Text;
+
+namespace SoapCore.Tests
+{
+	internal class MockMessageEncoder : System.ServiceModel.Channels.MessageEncoder
+	{
+		private bool _didWriteMessage = false;
+		public bool DidWriteMessage
+		{
+			get
+			{
+				return _didWriteMessage;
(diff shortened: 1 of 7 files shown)


tubs989:
Looks like we hit the same issue at the same time.

espcoder:
To use this implementation you have to register the service. If you do not
register the service it executes the same code path as before, so not sure
why it would not work in your case.

espcoder:
I figured out why this breaks you and have a fix in PR #222. It breaks because there is code to loop through all tuners and call tune. The fix is to not implement the IServiceOperationTuner interface. Sorry for the oversight on my part.

As I was looking into this issue I discovered another potential issue: AddSoapServiceOperationTuner tries to register a Singleton, which there can only be one, but then the middleware on line 282 loops through the registered IServiceOperationTuner(s) implying there can be more than one. Because it's a TryAddSingleton, the first one registered will be the one in the list.

tubs989:
> I figured out why this breaks you and have a fix in PR #222. It breaks because there is code to loop through all tuners and call tune. The fix is to not implement the IServiceOperationTuner interface. Sorry for the oversight on my part.
> 
> As I was looking into this issue I discovered another potential issue: AddSoapServiceOperationTuner tries to register a Singleton, which there can only be one, but then the middleware on line 282 loops through the registered IServiceOperationTuner(s) implying there can be more than one. Because it's a TryAddSingleton, the first one registered will be the one in the list.

I didn't actually use it I was saying that we encountered an issue where having a path with an action was breaking the flow. My pull request addressed a similar issue #219

PR #233, opened 2019-03-22
Fix #232 - support base types from different namespaces

### LINKED ISSUE #232: Support complex types with inheritance from different namespace
Given the following service setup:

```
namespace CoreService
{
    using System.ServiceModel;
    using Other;

    [ServiceContract]
    public interface ITestService
    {
        [OperationContract]
        Child GetChild(int id);
    }

    public class TestService : ITestService
    {
        public Child GetChild(int id)
        {
            return new Child();
        }
    }

    public class Child : Parent
    {
        public string Name { get; set; }
    }

}

namespace Other
{
    public class Parent
    {
        public int Id { get; set; }
    }
}
```

The current setup produces the following WSDL markup:

```
<xs:complexType xmlns:ser="http://schemas.microsoft.com/2003/10/Serialization/" name="Child">
<xs:complexContent>
<xs:extension base="Parent">
<xs:sequence>
<xs:element minOccurs="0" name="Name" nillable="true" type="xs:string"/>
</xs:sequence>
</xs:extension>
</xs:complexContent>
</xs:complexType>
```

We should aim for the following:

```
<xs:complexType xmlns:ser="http://schemas.microsoft.com/2003/10/Serialization/" name="Child">
<xs:complexContent>
<xs:extension xmlns:q1="http://schemas.datacontract.org/2004/07/Other" base="q1:Parent">
<xs:sequence>
<xs:element minOccurs="0" name="Name" nillable="true" type="xs:string"/>
</xs:sequence>
</xs:extension>
</xs:complexContent>
</xs:complexType>
```

I will be raising a PR to address this imminently

agpreynolds:


kotovaleksandr:
Thanks! Maybe should use GetDataContractNamespace instead GetModelNamespace? Base type can have DataContractAttribute with custom namespace

agpreynolds:
Good point, will change that now

Issue #243, opened 2019-04-08
XmlElement is not honored


cubidobusinesssolutions:
`XmlElementAttribute` is not working on array parameters

Code:
```cs
[ServiceContract(Namespace ="urn:test:1")]
public interface ITestService
{
    [OperationContract]
    TestResult ExecuteFunction([XmlElement("parameterArray")]string[] parameterArray);
}
```

This will parse an empty array.

cubidobusinesssolutions:
I will provide a PR for this and #180 in a minute

kotovaleksandr:
I trying to understand how XmlElement work and observe what array parameters with XmlSerializer never worked. 
In #248 user observe changed behavior of SoapCore, but i don't know how it works previously.   
I'm tried to modify your code for parsing array parameters and write some tests (look in https://github.com/DigDes/SoapCore/commit/cca43e19f9e166d8c5d7f08d4a0b964d497264c1). For case where array parameters used without any attributes tests passed and for first view works fine (exists problem with system element types such as string, int - which serialized in lower case tags, but in code types named with first upper case letter, fixed with check for System namespace).

But, i added to test service methods:
```csharp
[OperationContract]
[XmlSerializerFormat]
string[] PingStringArrayWithXmlArray([XmlElement("arrayItem")]string[] array);

[OperationContract]
[XmlSerializerFormat]
ComplexModel1[] PingComplexModelArrayWithXmlArray([XmlArrayItem("arr1")]ComplexModel1[] models, [XmlElement("arr2")]ComplexModel2[] models2);
``` 
And tests generates identical requests to server in both cases: with XmlArray (XmlElement) attribute and without them. Your code wait request body with different xml and test failed. Whats client you used for make requests to SoapCore?

If you can look to code and fix tests and SoapCore code (or write new correct tests) - i would thank you very much. I seem to be at a dead end.

cubidobusinesssolutions:
@kotovaleksandr I think this can be referenced (and closed) in latest [Release 0.9.9.5](https://github.com/DigDes/SoapCore/releases/tag/0.9.9.5), right?

kotovaleksandr:
No, this issue founded in 0.9.9.5 release, see original issue #248. Your PR breaks serialization of array responses.

cubidobusinesssolutions:
@kotovaleksandr Sry, I didn't see your comment until now
> Whats client you used for make requests to SoapCore?
[...]

Issue #180, opened 2019-01-13
XmlArrayItemAttribute is not honored


comburo:
I'm trying to port an existing .NET Framework WCF service to .NET Core using SOAP Core. Most of it seems to work as-is, which is great! But I just ran into an issue where the original service has an array of strings, but defines the name of the seperate items using an [XmlArryItem] attribute.

I've created a simple test service to illustrate the point.

Consider this Operation Contract:
``` C#
[ServiceContract(Namespace ="urn:test:1")]
public interface ITestService
{
    [OperationContract]
    TestResult ExecuteFunction(string firstParameter, [XmlArrayItem("item")]string[] parameterArray);
}
```

And the result object (not important, but for completeness):
``` C#
public class TestResult
{
    public string Message { get; set; }
}
```

This services uses the XmlSerializer:
``` C#
app.UseSoapEndpoint<TestService>("/Test.svc", new BasicHttpBinding(), SoapSerializer.XmlSerializer, false);
```

The WSDL will look like this (displaying only the relevant part):
``` XML
<xs:element name="ExecuteFunction">
    <xs:complexType>
        <xs:sequence>
            <xs:element minOccurs="0" maxOccurs="1" name="firstParameter" type="xs:string"/>
            <xs:element minOccurs="0" maxOccurs="1" name="parameterArray" type="tns:ArrayOfString"/>
        </xs:sequence>
    </xs:complexType>
</xs:element>
<xs:complexType name="ArrayOfString">
    <xs:sequence>
        <xs:element minOccurs="0" maxOccurs="unbounded" nillable="true" name="string" type="xs:string"/>
    </xs:sequence>
</xs:complexType>
<xs:element name="ArrayOfString" nillable="true" type="tns:ArrayOfString"/>
```

But with WCF, the WSDL would look like this (displaying only the array):
``` XML
<xs:complexType name="ArrayOfString">
    <xs:sequence>
        <xs:element minOccurs="0" maxOccurs="unbounded" nillable="true" name="item" type="xs:string"/>
    </xs:sequence>
</xs:complexType>
<xs:element name="ArrayOfString" nillable="true" type="tns:ArrayOfString"/>
```

Notice how the "name" of an individual array item is different.

I'll try to see if I can track down the issue, fix it and submit a PR, but I have no idea where to start looking. So any pointers would be appreciated :-)

kotovaleksandr:
Hi @comburo! By WCF you mean SoapSerializer.DataContractSerializer?
[...]

PR #244, opened 2019-04-08
  fix XmlSerialization of XmlElement and XmlArray array parameters and return types

cubidobusinesssolutions:
honor `XmlElement`, `XmlArray`, and `XmlArrayItem` attributes on array parameters at parsing via `XmlSerializer`

#180 
#243


kotovaleksandr:
Good job, thanks!

kotovaleksandr:
@cubidobusinesssolutions Please see #248. 
Your PR breaks default behavior of array parameters without XmlArray attribute. Test CheckArrays in https://github.com/DigDes/SoapCore/blob/develop/src/SoapCore.Tests/IntegrationTests.cs reproduce this issue. Can you fix them?

PR #258, opened 2019-05-17
Add custom exception transformer

jhaygood86:
WCF has the ability to define your own error handler (IErrorHandler) that formats an Exception to a SOAP Message.

This provides similar capability to SoapCore, using a simpler interface.

This also adds a helper class that many users of this interface will need (MessageFaultBodyWriter) since Microsoft in their infinite wisdom doesn't expose publically the API needed to take a MessageFault (created from a FaultException) and convert that to a Message.

This will allow services that used custom faults in WCF, but used an IErrorHandler to create them, to remain 100% API compatible when ported to SoapCore with minimal code changes.


kotovaleksandr:
Thanks! Can you add description to readme.md?

PR #260, opened 2019-05-23
Implement ability to use non-wrapped MessageContract

Simon-Campbell:
Hi,

MessageContract can have code wrappers that aren't represented in XML e.g.

```csharp
[MessageContract(IsWrapped = false)]
class FooRequest { 
   [MessageBodyMember(Order = 0)]
   public Foo Foo { get; set; }
}

class Foo {
    string FooValue { get; set; }
}
```

As I understand it, the XML that should work here is the following --

```
<Foo>
<FooValue>Test</FooValue>
</Foo>
```

Not:

```
<FooRequest>
<Foo>
<FooValue>Test</FooValue>
</Foo>
</FooRequest>
```

The wrapping complexity only manifests itself in C# code, not in the XML.

https://docs.microsoft.com/en-us/dotnet/framework/wcf/samples/unwrapped-messages

Update-- now does multiple parameters as per examples in linked MS WCF article

### DIFF
--- a/src/SoapCore.Tests/Serialization/Models.Xml/ISampleService.cs
+++ b/src/SoapCore.Tests/Serialization/Models.Xml/ISampleService.cs
@@ -54,6 +54,11 @@ namespace SoapCore.Tests.Serialization.Models.Xml
 		NotWrappedFieldComplexInputResponse NotWrappedFieldComplexInputRequestMethod(
 			NotWrappedFieldComplexInputRequest request);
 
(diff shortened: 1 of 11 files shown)


Simon-Campbell:
I decided to do the rejigging of GetRequestArguments to support multiple inlined parameters. The tests (including new ones) pass and my manual test of the new functionality still works

rinkeb:
@Simon-Campbell just wanted to let you know you know this solution fixed my wrapping issues, thanx! 

As far as i can see CoreWCF does not yet support this properly, so thats one good reason to keep using SoapCore over CoreWCF for now.

PR #278, opened 2019-06-19
Make a default FaultExceptionTransformer

Simon-Campbell:
## Summary
Pulling the existing fault logic which creates faults for unhandled exceptions into a `DefaultFaultExceptionTransformer`. 

This is useful if you want SOAP Core to figure out the "standard" SOAP fault logic, but want to decorate the response with your own non-standard things (without copying logic around)

### CustomFaultExceptionTransformer.cs
```csharp
    public class CustomFaultExceptionTransformer : IFaultExceptionTransformer
    {
        private readonly DefaultFaultExceptionTransformer _base;
        private readonly IHttpContextAccessor _httpContextAccessor;

        public CustomFaultExceptionTransformer(DefaultFaultExceptionTransformer @base, IHttpContextAccessor httpContextAccessor)
        {
            _httpContextAccessor = httpContextAccessor;
            _base = @base;
        }

        public Message ProvideFault(Exception exception, MessageVersion messageVersion)
        {
            var baseFault = _base.ProvideFault(exception, messageVersion);
            var httpContext = _httpContextAccessor.HttpContext;

            baseFault.Headers.Add(MessageHeader.CreateHeader("TraceIdentifier", "http://tempuri.org", httpContext?.TraceIdentifier ?? "None"));

            return baseFault;
        }
    }
```

### Startup.cs
```csharp
            services.AddHttpContextAccessor();
            services.AddScoped<DefaultFaultExceptionTransformer>();
            services.AddScoped<IFaultExceptionTransformer, CustomFaultExceptionTransformer>();
```

I am utilising it to add my own custom headers derived from the inbound request and took it as an opportunity to simplify the conditional code in SoapEndpointMiddleware.cs

### DIFF
--- a/src/SoapCore/DefaultFaultExceptionTransformer.cs
+++ b/src/SoapCore/DefaultFaultExceptionTransformer.cs
@@ -3,6 +3,10 @@ using System.ServiceModel.Channels;
 
 namespace SoapCore
 {
+	/// <summary>
+	/// The default implementation of the fault provider when an unexpected exception occurs. This can be replaced or
+	/// extended by registering your own IFaultExceptionTransformer in the service collection on startup.
+	/// </summary>
 	public class DefaultFaultExceptionTransformer : IFaultExceptionTransformer
 	{
 		private readonly ExceptionTransformer _exceptionTransformer;
--- a/src/SoapCore/DefaultFaultExceptionTransformer.cs
+++ b/src/SoapCore/DefaultFaultExceptionTransformer.cs
@@ -1,0 +1,31 @@
+using System;
+using System.ServiceModel.Channels;
+
+namespace SoapCore
+{
+	public class DefaultFaultExceptionTransformer : IFaultExceptionTransformer
+	{
+		private readonly ExceptionTransformer _exceptionTransformer;
+
+		public DefaultFaultExceptionTransformer()
+		{
+			_exceptionTransformer = null;
+		}
+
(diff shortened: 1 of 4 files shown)


henkmollema:
This change requires having a `IFaultExceptionTransformer` registered while it used to be optional.  For people running into this error after upgrading to 0.9.9.6:

> `InvalidOperationException`: No service for type `SoapCore.IFaultExceptionTransformer` has been registered.

You should add SoapCore services to your DI container in the startup class via:
```
services.AddSoapCore()`
```

Simon-Campbell:
> This change requires having a `IFaultExceptionTransformer` registered while it used to be optional. For people running into this error after upgrading to 0.9.9.6:
> 
> > `InvalidOperationException`: No service for type `SoapCore.IFaultExceptionTransformer` has been registered.
> 
> You should add SoapCore services to your DI container in the startup class via:
> 
> ```
> services.AddSoapCore();
> ```

Hmm. Sorry if that soaked up a lot of your time. Any better way I could have approached this? The docs say to call `services.AddSoapCore()` so that is the base constraint I tested with.

I'm thinking a better error message may have helped, but it's one of those things you only see in hindsight.

henkmollema:
I figured it out pretty quickly by searching for `IFaultExceptionTransformer` in this repo. I was surprised that I did not call `AddSoapCore`, simply forgot to do so and SoapCore happened to work without it until now. Since it's mentioned in the docs I guess not many people will stumble upon this. In case they do, I hope Google leads them to this PR. 😄

PR #286, opened 2019-07-16
Return empty body with void return (Help Needed)

Simon-Campbell:
## Summary
This change makes it so that an empty SOAP body (no generated wrapper) is returned with a void method e.g.

```
<s:Envelope xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:s="http://www.w3.org/2003/05/soap-envelope">
  <s:Body />
</s:Envelope>
```

The current behaviour is to add an unnecessary response envelope which is derived from the operation name e.g.

```
<s:Envelope xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:s="http://www.w3.org/2003/05/soap-envelope">
  <s:Body>
    <FooResponse xmlns="http://foo.namespace" />
  </s:Body>
</s:Envelope>
```

In my particular application use-case there are some additional SOAP headers added to all messages, but there is no content in the body.

## .NET Core WCF Client (Help)
I am having some issues getting the .NET Core WCF client to work with empty responses. It is used in the tests but is not happy about empty messages.  

What am I missing in able to get this to work? Is it a bug in that library or is something wrong with what I am trying to achieve?

This is the stack trace details from one of the skipped VoidAllMethod test in IntegrationTests.cs:

```
System.ServiceModel.CommunicationException
  HResult=0x80131500
  Message=Error in deserializing body of reply message for operation 'VoidAll'. The OperationFormatter could not deserialize any information from the Message because the Message is empty (IsEmpty = true).
  Source=System.Private.ServiceModel
  StackTrace:
   at System.ServiceModel.Dispatcher.PrimitiveOperationFormatter.DeserializeReply(Message message, Object[] parameters)
   at System.ServiceModel.Dispatcher.ProxyOperationRuntime.AfterReply(ProxyRpc& rpc)
   at System.ServiceModel.Channels.ServiceChannel.HandleReply(ProxyOperationRuntime operation, ProxyRpc& rpc)
   at System.ServiceModel.Channels.ServiceChannel.Call(String action, Boolean oneway, ProxyOperationRuntime operation, Object[] ins, Object[] outs, TimeSpan timeout)
   at System.ServiceModel.Channels.ServiceChannelProxy.InvokeService(MethodCall methodCall, ProxyOperationRuntime operation)
   at System.ServiceModel.Channels.ServiceChannelProxy.Invoke(MethodInfo targetMethod, Object[] args)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Reflection.DispatchProxyGenerator.Invoke(Object[] args)
   at SoapCore.Tests.IntegrationTests.VoidAllMethod() in C:\Users\CampbellS.DRC\dev\SoapCore\src\SoapCore.Tests\IntegrationTests.cs:line 44

Inner Exception 1:
SerializationException: The OperationFormatter could not deserialize any information from the Message because the Message is empty (IsEmpty = true).
```

### DIFF
--- a/src/SoapCore.Tests/ITestService.cs
+++ b/src/SoapCore.Tests/ITestService.cs
@@ -11,6 +11,12 @@ namespace SoapCore.Tests
 		[OperationContract]
 		string Ping(string s);
 
+		[OperationContract(IsOneWay = false)] // explicitly IsOneWay = false (synchronous)
+		void VoidAll();
+
+		[OperationContract(IsOneWay = false)] // explicitly IsOneWay = false (synchronous)
+		void VoidString(string s);
+
 		[OperationContract]
 		string EmptyArgs();
 
--- a/src/SoapCore.Tests/IntegrationTests.cs
+++ b/src/SoapCore.Tests/IntegrationTests.cs
@@ -41,8 +41,6 @@ namespace SoapCore.Tests
 
 			var client = CreateSoap12Client();
 
-			// We don't actually want an exception to be thrown but are asserting so
-			// that the behaviour change can be monitored.
 			client.VoidAll();
 		}
 
--- a/src/SoapCore.Tests/IntegrationTests.cs
+++ b/src/SoapCore.Tests/IntegrationTests.cs
@@ -35,6 +35,28 @@ namespace SoapCore.Tests
 		}
 
 		[TestMethod]
+		public void VoidAllMethod()
+		{
+			Assert.Inconclusive("The .NET WCF client does not appear to support empty messages");
+
+			var client = CreateSoap12Client();
+
+			// We don't actually want an exception to be thrown but are asserting so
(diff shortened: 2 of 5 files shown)


kotovaleksandr:
Hi @Simon-Campbell! I don't know how WCF client work internally. I can only offer you to create an issue in the WCF project. Please, fix merge conflicts so i can confirm PR.

Simon-Campbell:
@kotovaleksandr The merge conflicts have been resolved 👍

Simon-Campbell:
After further inspection, I will close this PR as the functionality required should be able to be achieved with a MessageInspector.

PR #288, opened 2019-07-17
Custom HttpResponse support

pengweiqhca:


kotovaleksandr:
Whats difference with IMessageInspector?

pengweiqhca:
This has nothing with `IMessageInspector`.

https://stackoverflow.com/questions/23440024/wcf-extension-for-adding-outgoing-message-properties-on-client-side-dynamically

Simon-Campbell:
> This has nothing with `IMessageInspector`.
> 
> https://stackoverflow.com/questions/23440024/wcf-extension-for-adding-outgoing-message-properties-on-client-side-dynamically

Not sure I understand how this is intended to be used, that link suggests you want to add custom headers to the client messages. Given that this is a server project, I assume you want the same functionality in server responses. 

This is achievable using an IMessageInspector + IFaultExceptionTransformer. Admittedly there is a bit of clunkyness (I'm not sure if faults should pass to IMessageInspector?). Here's a code sample of something similar to what I am using to achieve this in my own project.

```
public class Startup 
{
    public void ConfigureServices(IServiceCollection services)
    {
        services.AddScoped<IMessageInspector, CustomMessageInspector>();

        services.AddSingleton<DefaultFaultExceptionTransformer>();
        services.AddScoped<IFaultExceptionTransformer, CustomFaultHeaderFaultExceptionTransformer>();
    }
}
```

```
// Add headers to non-faults
public class CustomMessageInspector : IMessageInspector
{
    public object AfterReceiveRequest(ref Message message)
    {
        return null;
    }

    public void BeforeSendReply(ref Message reply, object correlationState)
    {
        if (reply.IsFault)
        {
            // SOAP Core does not call message inspectors with for fault (might be a bug). Let's explicitly capture this
            // and return early. We can service the error use case with IFaultExceptionTransformer
            return;
        }


        var headers = message.Headers;

        headers.Add(ChannelMessageHeader.CreateHeader("Hello", "http://tempuri.org", "World"));
    }
}
```

```
// Adds headers to faults
public class CustomFaultHeaderFaultExceptionTransformer : IFaultExceptionTransformer
{
    private readonly DefaultFaultExceptionTransformer _defaultFaultExceptionTransformer;

    public DigadFaultExceptionTransformer(DefaultFaultExceptionTransformer defaultFaultExceptionTransformer)
    {
        _defaultFaultExceptionTransformer = defaultFaultExceptionTransformer;
    }

    public Message ProvideFault(Exception exception, MessageVersion messageVersion)
    {
        var baseMessage = _defaultFaultExceptionTransformer.ProvideFault(exception, messageVersion);        
        var headers = baseMessage.Headers;

        headers.Add(ChannelMessageHeader.CreateHeader("Hello", "http://tempuri.org", "World"));

        return baseMessage;
    }
}
```

src/SoapCore/SoapEndpointMiddleware.cs
```diff
@@ -57,7 +55,7 @@ public SoapEndpointMiddleware(ILogger<SoapEndpointMiddleware> logger, RequestDel
 
 		public async Task Invoke(HttpContext httpContext, IServiceProvider serviceProvider)
 		{
-			httpContext.Request.EnableRewind();
+			httpContext.Request.EnableBuffering();
```
Simon-Campbell: It looks like you have some bonus changes. I presume that's because you merged master into your branch instead of develop.

pengweiqhca: I has merged master

Simon-Campbell:
@pengweiqhca 

I have reviewed this PR but tidied up the odd merge behaviour from when you merged master. I have rebased your single commit on top of the current develop branch and validated that your tests still pass.

Please see #338

PR #303, opened 2019-08-03
Add support for circular references in datacontract objects

### LINKED ISSUE #302: Add support for circular references in data contracts
Hi!

First of all, thank you for your project.

It appears that circular references in datacontract objects are not well supported, and always lead to a StackOverflowException. This seems to be due to the way recursivity is handled while parsing objects in the `MetaWCFBodyBuilder` class.

Short example :
```csharp
[DataContract]
public class CircularReferenceFirstObject
{
    [DataMember]
    public CircularReferenceSecondObject SecondObject { get; set; }
}
```

references 
```csharp
[DataContract]
public class CircularReferenceSecondObject
{
    [DataMember]
    public CircularReferenceFirstObject FirstObject { get; set; }
}
```
which references `CircularReferenceFirstObject`. StackOverflow is raised on first call.

apatenotre:
Fixes #302 

Solution : store type in dictionary before the recursive call

### DIFF
--- a/src/SoapCore.Tests/Wsdl/Services/CircularReferenceFirstObject.cs
+++ b/src/SoapCore.Tests/Wsdl/Services/CircularReferenceFirstObject.cs
@@ -1,0 +1,11 @@
+using System.Runtime.Serialization;
+
+namespace SoapCore.Tests.Wsdl.Services
+{
+    [DataContract]
+    public class CircularReferenceFirstObject
+    {
+		[DataMember]
+        public CircularReferenceSecondObject SecondObject { get; set; }
+    }
+}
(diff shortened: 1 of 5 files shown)


kotovaleksandr:
Hi @apatenotre Please, fix conflicts. Thanks for your help!

apatenotre:
Hi @kotovaleksandr,
 work in progress, will fix this tonight and let you know!

apatenotre:
> 
> 
> Hi @apatenotre Please, fix conflicts. Thanks for your help!

All good

apatenotre:
Actually, moving it back to WIP as I think I found a regression. I will clean it tomorrow

apatenotre:
Ok, should be good now.
The idea is that tracking of objects discovered while building the wsdl was incorrectly done. `_complexTypeToBuild` was used to detect circular references, but was set even before the type was read for the first time. So types could be skipped when they were not root while discovered. So I introduced a dedicated hashset

Issue #304, opened 2019-08-03
Namespace on DataContractAttribute for a request argument results in parse issue


jrinker03:
This appears to be the same issue as the second issue mentioned in #229 and I am able to repro with my own test classes and server.

The easiest way to see this issue is to add a namespace to the DataContractAttribute of the test class ComplexModel2.  This will cause several unit tests to fail.  Like so:
```
	[DataContract(Namespace = "http://SoapCore.Tests.Serialization.Models.DataContract.ComplexModel2")]
	public class ComplexModel2
	{
```
The request looks like this:
```
<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/">
    <s:Body>
        <PingComplexModel xmlns="http://tempuri.org/">
            <inputModel xmlns:a="http://SoapCore.Tests.Serialization.Models.DataContract.ComplexModel2" xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
                <a:ComplexListProperty xmlns:b="http://schemas.datacontract.org/2004/07/SoapCore.Tests.Serialization.Models.DataContract">
                    <b:ComplexObject>
                        <b:IntProperty>1</b:IntProperty>
                        <b:StringProperty>ComplexObject sample one</b:StringProperty>
                    </b:ComplexObject>
                    <b:ComplexObject>
                        <b:IntProperty>2</b:IntProperty>
                        <b:StringProperty>ComplexObject sample two</b:StringProperty>
                    </b:ComplexObject>
                </a:ComplexListProperty>
                <a:ComplexNestedObjectProperty xmlns:b="http://schemas.datacontract.org/2004/07/SoapCore.Tests.Serialization.Models.DataContract">
                    <b:IntProperty>2</b:IntProperty>
                    <b:StringProperty>ComplexObject sample two</b:StringProperty>
                </a:ComplexNestedObjectProperty>
                <a:EnumProperty>B</a:EnumProperty>
                <a:FloatProperty>2.22</a:FloatProperty>
                <a:IntProperty>22</a:IntProperty>
                <a:ListProperty xmlns:b="http://schemas.microsoft.com/2003/10/Serialization/Arrays">
                    <b:string>one2</b:string>
                    <b:string>two2</b:string>
                </a:ListProperty>
                <a:StringProperty>ComplexModel2 sample two</a:StringProperty>
            </inputModel>
        </PingComplexModel>
    </s:Body>
</s:Envelope>
```
[...]

PR #305, opened 2019-08-03
Fix null issue when there is a namespace on DataContractAttribute for a request argument

jrinker03:
This is the fix for issue https://github.com/DigDes/SoapCore/issues/304

The namespace being passed into DeserializeInputParameter is the namespace of the members of the input rather than the input object itself.

This change fixes the above and adds a namespace to the unit test class ComplexModel2 to catch this in the fuiture.

### DIFF
--- a/src/SoapCore.Tests/Serialization/Models.DataContract/ComplexModel2.cs
+++ b/src/SoapCore.Tests/Serialization/Models.DataContract/ComplexModel2.cs
@@ -4,7 +4,7 @@ using System.Runtime.Serialization;
 namespace SoapCore.Tests.Serialization.Models.DataContract
 {
 	// same as ComplexModel1, just for testing
-	[DataContract]
+	[DataContract(Namespace = "http://SoapCore.Tests.Serialization.Models.DataContract.ComplexModel2")]
 	public class ComplexModel2
 	{
 		[DataMember]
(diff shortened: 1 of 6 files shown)


src/SoapCore.Tests/Serialization/Models.DataContract/ComplexModel2.cs
```diff
@@ -4,7 +4,7 @@
 namespace SoapCore.Tests.Serialization.Models.DataContract
 {
 	// same as ComplexModel1, just for testing
-	[DataContract]
+	[DataContract(Namespace = "http://SoapCore.Tests.Serialization.Models.DataContract.ComplexModel2")]
```
Simon-Campbell: This might unexpectedly change test coverage. What happens if in the scenario where the namespace is left undefined?

jrinker03: No change (error) would happen to the input model with no namespace but I've added a test case to prove and enforce this.

jrinker03:
@Simon-Campbell, any chance you can take a look and provide any additional feedback?

Simon-Campbell:
> @Simon-Campbell, any chance you can take a look and provide any additional feedback?

Hi, I have had another quick look. All existing tests seem to pass with no reduced coverage. 

The main gotcha appears to be captured, that is: you identified the bug for DataContractSerializer so the fix may have broken XmlSerializer. 

You have coverage for both serializers 👍

I do find it peculiar that in the old experssion `parameterInfo.Namespace ?? operation.Contract.Namespace` that parameterInfo.NameSpace did not have the correct value. 

I did not write the original code, and there is not existing comment as to why there is fallback so I would be guessing

jrinker03:
Thanks @Simon-Campbell!    I was a bit surprised about the parameterInfo.Namespace as well but testing proved out that the input object's namespace always has the parent elemement's namespace.

@kotovaleksandr, any thoughts on this change?

PR #310, opened 2019-08-07
XmlSerializer: Implement MessageContract(IsWrapped=false) for return results

Simon-Campbell:
This follows up the PR where I implemented IsWrapped=false for input parameters (#260). This PR implements it for return parameters. It follows the documentation outlined here.

https://docs.microsoft.com/en-us/dotnet/framework/wcf/samples/unwrapped-messages

### DIFF
--- a/src/SoapCore.Tests/RawRequestSoap12Tests.cs
+++ b/src/SoapCore.Tests/RawRequestSoap12Tests.cs
@@ -47,6 +47,33 @@ namespace SoapCore.Tests
 		}
 
 		[TestMethod]
+		public void Soap12PingWithActionInEnvelopeHeader()
+		{
+			var body = @"<?xml version=""1.0"" encoding=""utf-8""?>
+<soap12:Envelope xmlns:xsi=""http://www.w3.org/2001/XMLSchema-instance"" xmlns:xsd=""http://www.w3.org/2001/XMLSchema"" xmlns:soap12=""http://www.w3.org/2003/05/soap-envelope"" xmlns:wsa=""http://www.w3.org/2005/08/addressing"">
+  <soap12:Header>
+	<wsa:Action>Ping</wsa:Action>
+  </soap12:Header>
+  <soap12:Body>
+    <Ping xmlns=""http://tempuri.org/"">
+      <s>abc</s>
+    </Ping>
+  </soap12:Body>
+</soap12:Envelope>
+";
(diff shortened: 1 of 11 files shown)


src/SoapCore.Tests/RawRequestSoap12Tests.cs
```diff
@@ -46,6 +46,33 @@ public void Soap12PingWithActionInHeader()
 			}
 		}
 
+		[TestMethod]
+		public void Soap12PingWithActionInEnvelopeHeader()
```
Simon-Campbell: This test is unrelated but to avoid the admin overhead. I have included it to increase test coverage.

src/SoapCore/ServiceBodyWriter.cs
```diff
@@ -20,151 +22,173 @@ public class ServiceBodyWriter : BodyWriter
 		private readonly object _result;
 		private readonly Dictionary<string, object> _outResults;
 
-		public ServiceBodyWriter(SoapSerializer serializer, OperationDescription operation, string resultName, object result, Dictionary<string, object> outResults) : base(isBuffered: true)
```
Simon-Campbell: I followed this parameter and noted it was just lifted out of operation.ReturnName. I have pulled that logic in here and removed the extra parameter.

src/SoapCore/ServiceBodyWriter.cs
```diff
@@ -20,151 +22,173 @@ public class ServiceBodyWriter : BodyWriter
...
 			_result = result;
-			_outResults = outResults;
+			_outResults = outResults ?? new Dictionary<string, object>();
 		}
 
 		protected override void OnWriteBodyContents(XmlDictionaryWriter writer)
 		{
-			// do not wrap old-style single element response into additional xml element for xml serializer
-			var needResponseEnvelope = _serializer != SoapSerializer.XmlSerializer || _result == null || (_outResults != null && _outResults.Count > 0) || !_operation.IsMessageContractResponse;
+			switch (_serializer)
```
Simon-Campbell: Instead of checking the cases in-line I have opted to split the xml/data contract logic internally at a higher level.

This has resulted in simpler logic to follow if you are concerned about a single path and allows me to isolate the serialization change (to not unexpected impact DataContractSerializer).

Unfortunately it does duplicate some logic.

Simon-Campbell:
Hi @kotovaleksandr - are you able to publish a new alpha build which includes this? 👍

PR #314, opened 2019-08-11
Handle nullable enums

### LINKED ISSUE #313: Nullable enums are considered as nullable ints
In case of nullable enums, underlying type should be correctly taken into account. At the moment, it is not the case and instead, are considered as nullable ints

apatenotre:
Fixes #313

### DIFF
--- a/src/SoapCore.Tests/Wsdl/Services/INonNullableEnumService.cs
+++ b/src/SoapCore.Tests/Wsdl/Services/INonNullableEnumService.cs
@@ -1,0 +1,20 @@
+using System;
+using System.ServiceModel;
+
+namespace SoapCore.Tests.Wsdl.Services
+{
+	[ServiceContract]
+	public interface INonNullableEnumService
+	{
+		[OperationContract]
+		NulEnum GetEnum(string text);
+	}
+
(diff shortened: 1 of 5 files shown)


kotovaleksandr:
Hi, thanks! Please, add tests for this case. Positive tests for nullable and non-nullable enums should be enough.

apatenotre:
Hey, yes with pleasure, however I am still unsure how to do this. 
The best way would be to dynamically generate client classes based on wsdl definition, and ensure properties are correctly created with nullable enums, but this is not trivial. 

I will keep thinking about it. If you have any idea, please let me know!

kotovaleksandr:
I know, is not simple. But you can analyze wsdl as xml. Tests contains service with test contract, you can use them. This will guarantee that your improvements will not be removed by someone in the future.

apatenotre:
I can do it that way, yes. I will see how to make it a bit sustainable

apatenotre:
Hi @kotovaleksandr , so here is a test attempt - let me know if this suits you

The basic idea is to parse the WSDL as an XML to ensure that expected types are generated
* one complex type and one simple type when a nullable enum is used somewhere
* one simple type only when no nullable enum is used

PR #321, opened 2019-08-20
[WIP] Configure get service metadata

### LINKED ISSUE #320: Add possibility to forbid access to WSDL
The idea is to implement the following property, currently available in a WCF project
* https://docs.microsoft.com/fr-fr/dotnet/api/system.servicemodel.description.servicemetadatabehavior.httpgetenabled?view=netframework-4.8

This allows server to reject GET request on WSDL, and is useful when the service contract should be hidden from the user

apatenotre:
Fixes #320 

Still in progress :)

### DIFF
--- a/src/SoapCore.Tests/TrailingServicePathTuner/TrailingServicePathTunerTests.cs
+++ b/src/SoapCore.Tests/TrailingServicePathTuner/TrailingServicePathTunerTests.cs
@@ -36,6 +36,7 @@ namespace SoapCore.Tests
 
 			var context = new DefaultHttpContext();
 			context.Request.Path = new PathString("/DynamicPath/Service.svc");
+			context.Request.Method = "GET";
 
 			// Act
 			// MockServiceProvider(false) simulates registering the TrailingServicePathTuner in app startup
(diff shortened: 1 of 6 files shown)


apatenotre:
Ok, I have been busy lately so could not take some time to finish this one. 
Now it should be fine as a first proposal.

I made the choice to add a new configuration API :
```csharp
public static IApplicationBuilder UseSoapEndpoint<T>(this IApplicationBuilder builder, Action<SoapCoreOptions> options);
```

The idea is to use an ASP.NET Core-style pattern, with an Action to be provided by the consumer of the API. It embeds all parameters that are configurable in the already existing overloads.

My advice would be to deprecate the other APIs and leave this one alone, which is written in a more modern style , and allows configuration of all parameters.

I can create another PR for this change if you like it

apatenotre:
Also I should write a few tests. Do you have any preference on where to put them in the Tests project ?

apatenotre:
Hi @kotovaleksandr , do you have any feedback regarding this one ?
Let me know if I can help/improve!

Issue #322, opened 2019-08-21
Sample code build error with incompatible packages.


jrinker03:
When taking a fresh copy of the repository and building in VS 2019, the following error happens and the build fails:

```
5>D:\src\jrinker03\SoapCore\samples\Client\Client.csproj : error NU1605: Detected package downgrade: System.ServiceModel.Http from 4.5.3 to 4.4.0. Reference the package directly from the project to select a different version. 
5>D:\src\jrinker03\SoapCore\samples\Client\Client.csproj : error NU1605:  Client -> Server -> SoapCore -> System.ServiceModel.Http (>= 4.5.3) 
5>D:\src\jrinker03\SoapCore\samples\Client\Client.csproj : error NU1605:  Client -> System.ServiceModel.Http (>= 4.4.0)
```

The fix here is to update the sample projects to use System.ServiceModel.* v4.5.3 which is both the latest RTM editions and what SoapCore is compiled with.

Issue #316, opened 2019-08-14
Please update the sample, Complex object doesn't work


OscarCanek:
I cloned the repo but when running the complex object action always received null properties at the server.

OscarCanek:
Now is working, Thank you 👍

PR #323, opened 2019-08-21
[Samples] Fix compiler error with incompatible packages and ComplexModelResponse not being deserialized.

jrinker03:
Fix compiler error with incompatible packages (#322) and ComplexModelResponse not being deserialized (#316)

Also add output of the response object's enum property.


kotovaleksandr:
Thanks!

Issue #326, opened 2019-09-11
MessageBodyMemberAttribute.Name is being ignored


d4nuu8:
I am having a class like this

```csharp
[MessageContract]
public sealed class Request
{
    [MessageBodyMember(Name = "inputXML")]
    public string InputXml { get; set; }
}
```

In the generated WSDL I would expect

```xml
<xs:complexType name="Request">
    <xs:sequence>
        <xs:element minOccurs="0" maxOccurs="1" name="inputXML" type="xs:string"/>
    </xs:sequence>
</xs:complexType>
```

but I got

```xml
<xs:complexType name="Request">
    <xs:sequence>
        <xs:element minOccurs="0" maxOccurs="1" name="InputXml" type="xs:string"/>
    </xs:sequence>
</xs:complexType>
```

It seems like the MessageBodyAttribute.Name is completely ignored.

The SOAP endpoint is created like this

```csharp
app.UseSoapEndpoint<...>("/service.asmx", new BasicHttpBinding(), SoapSerializer.XmlSerializer)
```
Currently I am using version 0.9.9.6.

PR #328, opened 2019-09-12
[WIP] Fix for issue #326: MessageBodyMemberAttribute.Name is being ignored.

paleocomburo:
I have made a simple addition to MessageBodyMemberAttribute is handled always if the parent object has a MessageContractAttribute. Before it was only handled if that attribute was set to "IsWrapped = false".

I have manually verified it works, but, as per contributing guidelines, I want to add relevent unit-tests. But I have no idea where to start. I saw existing serialization tests, but these are not helpful because this is about the XML being generated. A class can perfectly serialize data wrong and deserialze that same wrong output back again. We need to know the serialization is not wrong.
The raw SOAP 1.2 tests seem more like the thing I need, but then I would have to write the necessary plumbing to make those tests possible. Just adding them to the existing SOAP 1.2 tests seem incorrect as this has nothing the to with SOAP 1.2 specifically.

Any guidance?

### DIFF
--- a/src/SoapCore/ServiceBodyWriter.cs
+++ b/src/SoapCore/ServiceBodyWriter.cs
@@ -148,13 +148,15 @@ namespace SoapCore
 				else
 				{
 					var messageContractAttribute = resultType.GetCustomAttribute<MessageContractAttribute>();
-
-					// This behavior is opt-in i.e. you have to explicitly have a [MessageContract(IsWrapped=false)]
-					// to have the message body members inlined.
-					var shouldInline = messageContractAttribute != null && messageContractAttribute.IsWrapped == false;
-
-					if (shouldInline)
+					if (messageContractAttribute != null)
 					{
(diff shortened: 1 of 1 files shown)


Simon-Campbell:
Adding to the SOAP 1.2 tests is probably good enough coverage to start with. 

IMO tackle it in that way first if you don't want to duplicate the plumbing, alternatively you could look at making the plumbing better configurable. 

I have no specific suggestions there but would be happy to merge if you added to the SOAP 1.2 scenario.

A question I have after consulting [this](
https://docs.microsoft.com/en-us/dotnet/framework/wcf/feature-details/using-message-contracts) (I have not had the opportunity to pull the code and review in full yet) --

Are you planning on implementing support for WrapperName and WrapperNamespace?

> The MessageContractAttribute allows you to specify the WrapperName and WrapperNamespace attributes which control the name of the wrapper element in the body of the SOAP message. **By default the name of the message contract type is used for the wrapper and the namespace in which the message contract is defined http://tempuri.org/ is used as the default namespace.**

PR #349, opened 2019-10-28
Fix/wcf wsdl non wrapped elements

jhaygood86:
When a service uses the XmlSerializer, but uses non-wrapped elements, the WSDL generation is horribly wrong.

This fixes that to properly unwrap elements when svcutil.exe generates such contracts.

Note: the operation already does the wrapping/unwrapping properly from what I can see. The WSDL generated however wouldn't work.


kotovaleksandr:
Thanks! Can you add tests for this change in future?

jhaygood86:
Will do. Was just making sure I had the PR created before I forgot to 😀

gorillapower:
Hello, do you know when this PR will be released? We are encountering this issue when we upgraded to netcore 3.1

jhaygood86:
This was merged ages ago, so it should be in current nuget packages

PR #423, opened 2020-02-03
Add support for DataSet/DataTable and any other type  that has XmlSchemaProviderAttribute

EamonHetherton:
...this includes removing specific code to support System.Xml.Linq.XElement which also happens to have the XmlSchemaProviderAttribute on it.

Would like to do a bigger refactor on the MetaBodyWriter and MetaWCFBodyWriter to use the XmlSchema objects rather than raw xmlWriter instructions as time permits....

### DIFF
--- a/src/SoapCore.Tests/Wsdl/Services/IDatetimeOffsetService.cs
+++ b/src/SoapCore.Tests/Wsdl/Services/IDatetimeOffsetService.cs
@@ -1,0 +1,20 @@
+using System;
+using System.ServiceModel;
+
+namespace SoapCore.Tests.Wsdl.Services
+{
+	[ServiceContract]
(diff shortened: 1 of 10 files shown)


EamonHetherton:
would love another alpha nuget package including these changes :)

Not sure how much work is involved but would be nice to have a pre-release nuget package automatically built whenever a pull request is made.  Have had a very positive experience with some Serilog libraries where my pull requests automatically create a pre-release nuget package if the build passes and I could use it right away, very nice experience and encourages contributions :)

kotovaleksandr:
@EamonHetherton https://www.nuget.org/packages/SoapCore/1.0.0.2-alpha
Now nuget package published on push to master branch. I will try to change github action script to publish on each merge to develop branch, may be on this week:)

PR #427, opened 2020-02-06
WIP: Add support for overriding namespace prefixes and handle XmlType attribute

EamonHetherton:
Moving towards a more strongly typed Xml creation for meta data.  Cleans up the namespace code a bit before I can better refactor the hand rolled XmlWriter statements into a more robust implementation using XmlSchema types.

With this commit, my new soap service generates near identical WSDL to an old WebService implementation, significantly closer than it used to.


kotovaleksandr:
@EamonHetherton Hi! Finished? I can merge?:)

EamonHetherton:
I'm using it in production, so I think yes.  More to do later as time permits but I'll create a new pull request then .

PR #429, opened 2020-02-10
Support for ISO-8859-1

microknights:
Today SoapCore only supports encoding of `utf-8, utf-16 (+le/be)`, and not any legacy like `ISO-8859-1`.

So when starting the SoapCore consuming part like this:
```
            var soapEncoder = new SoapEncoderOptions
            { 
                WriteEncoding = Encoding.GetEncoding("ISO8859-1"),
                MessageVersion = MessageVersion.Soap11
            };
            app.UseSoapEndpoint<IMySoapConsumer>("/my/soap/consumer", soapEncoder,  SoapSerializer.XmlSerializer);
```
Will end in a exception, since `XmlDictionaryReader.CreateTextReader` only support the mentioned encodings.

This PR makes it so it falls back to `XmlReader` which has no encoding restrictions, and thereby support ISO-8859-1.

### DIFF
--- a/src/SoapCore.Tests/IntegrationTests.cs
+++ b/src/SoapCore.Tests/IntegrationTests.cs
@@ -1,10 +1,12 @@
 using System;
 using System.ServiceModel;
 using System.ServiceModel.Channels;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting;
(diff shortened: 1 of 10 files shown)


kotovaleksandr:
Hi @microknights! Thanks for your contributing! Can you add unit tests for this change?

PR #460, opened 2020-04-17
Passed the right encoder to WriteErrorResponseMessage to handle Soap1…

hircjusz:
Passed the right encoder to WriteErrorResponseMessage to handle Soap1.1 format and 1.2 format.
I had problem with Fault message handling using endpoint with Soap1.1 and Soap1.2 because it was passed always  first encoder without matching the Message Version so  we always get FaultException in one version of SOAP.  Can it be reviewed and merged asap and some nugget published? I need this change in project


kotovaleksandr:
Thanks! Package already pushed to nuget, please check.

PR #468, opened 2020-04-27
Passed request message to FaultException provider

hircjusz:
It's is useful to have original message in provide FaultExceptionProvider. For example we can find out what method of WS was invoked from requestMessage Headers. It's very useful, otherwise we need to iterate stack trace of exception


hircjusz:
Please publish new version of nuget

kotovaleksandr:
@hircjusz https://www.nuget.org/packages/SoapCore/1.1.0-alpha
Thanks for contributing!

Issue #401, opened 2020-01-09
DataContract containing Array[] produces invalid WSDL


NikErmRnd:
Hi, I was faced with the problem of incorrect generation of wsdl, it consists in the fact that when creating an array, an unnecessary subobject is generated that interferes with the validation of the target system. For verification, when creating an ASMX system in legacy, this does not happen.

Fox example:

Model:
```
    [XmlRoot]
    public class BusinessMessages
    {
        [XmlElement("Objects", Form = System.Xml.Schema.XmlSchemaForm.Unqualified)]
        public BusinessMessage[] Objects { get; set; }
    }
```

```
    [XmlRoot("Objects")]
    public class BusinessMessage
    {
        public int Id { get; set; }
        public string Message { get; set; }
    }
```

How to remove this invalid subobject? "BusinessMessages"

And I got 
With SoapCore: 
![image](https://user-images.githubusercontent.com/22419495/72076715-0cc13e00-3307-11ea-9725-0790d12a2f59.png)

With Legacy ASMX template (.net framework 4.7):
![image](https://user-images.githubusercontent.com/22419495/72075853-93751b80-3305-11ea-9e30-ab4a49b7b1f3.png)

PR #469, opened 2020-04-27
Fix DataContract containing Array[] produces invalid WSDL

NikErmRnd:
https://github.com/DigDes/SoapCore/issues/401

Current realization not provide array by wsdl standard. I found this earlies and sent examples.

Now this is correct.


kotovaleksandr:
Hi @NikErmRnd! Please, add unit test for this case.

NikErmRnd:
> Hi @NikErmRnd! Please, add unit test for this case.
@kotovaleksandr 
Hi, test was added

PR #472, opened 2020-05-02
Bugfix for nullable value types in wsdl

oruchreis:
Hi,
Nullable non-system structs and enums are generiting as "Nullable`1" in wsdl with XmlSerializer. I've fixed these situations in this PR. It fixes also wrong ns attribute for message faults in MetaWCFBodyWriter.
Best regards.


kotovaleksandr:
Hi @oruchreis! Please add test for this change.

PR #476, opened 2020-05-13
Support for mixing serializers

Cornishman:
Due to a scenario where I needed the flexibility of XML processing incoming / outgoing requests but needed the meta (WSDL) endpoint to generate using the DataContract serializer.

This PR add's the ability to allow a user to override the serializer being used for the meta endpoint, by default it's not set so will use the existing SoapSerializer setting.


andersjonsson:
Closing this PR. Please resolve conflicts and re-open if still relevant

PR #496, opened 2020-07-05
Improvement of generic type support and Dictionary handling

simonsoanes:
When a service contains Dictionary<string, string> or any other complex type parameter, only the first generic parameter is included and turned into an array, rendering the dictionary as string[] in the proxy.

-- 

This adds fixes to the WSDL generation to support messages containing generic types with more than one generic parameter and support for ICollection in preference over IEnumerable so that collection types are known.  It also handles String being an IEnumerable but needing to be special cased to String in the WSDL.

This fixes WSDL generation so that the parser built into Visual Studio can handle generic collections properly and dictionaries.  It was unable to find types previously as the generated WSDL was malformed and had the wrong names and/or namespaces.

There's some unknowns around how the original WCF parser special cases dictionaries still. Currently, it represents it as a list of KeyValuePairTypeType in the generated proxy, I've made SoapCore behave very close to WCF's original mechanism but it doesn't seem to detect it as a Dictionary yet, when I find how it does it I'll submit another PR to align perfectly to WCF.

### DIFF
--- a/src/SoapCore.Tests/ITestService.cs
+++ b/src/SoapCore.Tests/ITestService.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.ServiceModel;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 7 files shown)


simonsoanes:
This may also fix the namespace issue someone else is reporting (if someone else hasn't), as it picks the correct type instead of using the wrong one, I can see it's using the generics namespace correctly for example.

The tests all pass locally, I'll try to recreate the problem.

simonsoanes:
I've also had to fix the underlying start service method.  It had a race condition in it, where the Kestrel server instance could be partially started when looking at the server addresses on a slower container and it would fail tests at random as the collection was empty.

PR #524, opened 2020-09-24
1.1.0.2

kotovaleksandr:
Updated with exception handeling when generating message #523


kotovaleksandr:
@nc-sosc Thank you! Please: https://www.nuget.org/packages/SoapCore/1.1.0.2

PR #538, opened 2020-10-27
fix #319 by switching serialization order of return param and out params

### LINKED ISSUE #319: Serialization order of return value vs ref parameters seems wrong
When a function both has a result as out/ref parameters the response appears to be serialized in the wrong order, in any case it is not possible to consume the services from a regular dotnet client.

e.g.
[OperationContract]
TBRSFlightData[] GetTodaysFlightsByType(ref string SecurityToken);

response gets serialized like this:

\<GetTodaysFlightsByTypeResponse xmlns="http://tempuri.org/">
    \<SecurityToken>xxxxxxx\</SecurityToken>
    \<GetTodaysFlightsByTypeResult .....
     \<a:TBRSFlightData>
        ........
      \</a:TBRSFlightData>
    \</GetTodaysFlightsByTypeResult>
\</GetTodaysFlightsByTypeResponse>"

but it should be like this:

\<GetTodaysFlightsByTypeResponse xmlns="http://tempuri.org/">
   \<GetTodaysFlightsByTypeResult .....
      \<a:TBRSFlightData>
      ........
      \</a:TBRSFlightData>
    \</GetTodaysFlightsByTypeResult>
    \<SecurityToken>xxxxxxx</SecurityToken>
\</GetTodaysFlightsByTypeResponse>

After creating a fork and changing the serialization order I am able to consume the service with our existing client.

DominicUllmann:
WCF client can only deserialize out/ref parameter together with return parameter, if serialized in the correct order. 
Changing result and out/ref parameter order is necessary so that ref/out parameters are not deserialized as null.

This will fix the issue #319

### DIFF
--- a/src/SoapCore/ServiceBodyWriter.cs
+++ b/src/SoapCore/ServiceBodyWriter.cs
@@ -89,61 +89,6 @@ namespace SoapCore
 				writer.WriteStartElement(_envelopeName, _serviceNamespace);
 			}
 
-			foreach (var outResult in _outResults)
-			{
-				string value = null;
-				if (outResult.Value is Guid)
-				{
-					value = outResult.Value.ToString();
-				}
-				else if (outResult.Value is bool)
-				{
-					value = outResult.Value.ToString().ToLower();
-				}
-				else if (outResult.Value is string)
-				{
-					value = System.Security.SecurityElement.Escape(outResult.Value.ToString());
-				}
-				else if (outResult.Value is Enum)
-				{
-					value = outResult.Value.ToString();
-				}
-				else if (outResult.Value == null)
-				{
-					value = null;
-				}
-				else
-				{
-					//for complex types
-					using (var ms = new MemoryStream())
-					using (var stream = new BufferedStream(ms))
-					{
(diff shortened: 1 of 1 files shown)


elca-lan (approved):
This is essentially the same change that @lokasoft already did in his fork and I think it makes sense like this because the old way is simply not conforming to the spec as far as I can tell.

StevenBonePgh:
Would it be possible to push a new package with this change?  Thanks!

kotovaleksandr:
@StevenBonePgh @elca-lan https://www.nuget.org/packages/SoapCore/1.1.0.4-beta

DominicUllmann:
Great, thanks a lot for the new release. It works well 👍

sadollah1:
I install the beta version but ı still need the change order number manually.The problem is still not solved for me. Any advice.

DominicUllmann:
> 
> 
> I install the beta version but ı still need the change order number manually.The problem is still not solved for me. Any advice.

How did you test that? It works perfectly for me with a wcf client. How does your service interface and client look like? What do you mean by changing the order manually?

Issue #551, opened 2020-12-03
Feature to add non generic UseSoapEndpoint extension api for use with reflection at runtime


dwigand:
I'm currently working on a soap api using a plugin architectur to register multiple SoapEndpoints at startup.

Therefor i need a UseSoapEndpoint(this IApplicationBuilder builder, Action<SoapCoreOptions> opt, Type serviceType) extenion api to set the serviceType at runtime and have full control over the SoapOptions.

I'll create a pr to accomplish this.

PR #552, opened 2020-12-04
Add additional signatures to SoapEndpointExtensions

dwigand:
Can be used to support non generic usage within an plugin architecture that determines the serviceTypes at startup.


dwigand:
This PR can be linked to Issue #551

Issue #543, opened 2020-11-11
Allow passing parameters in different order


purpurproger:
When we use SoapCore we should place parameters in a soap message the same order as it defined in service.

For example, if service contains method
```
List<Post> GetPosts(int offset, int count)
```

Client application should sends message like that
```
<soap:Body>
	<GetPosts>
		<offset>100</offset>
		<count >50</count >
	</GetPosts>
</soap:Body>
```

but if it sends the following message the parameters won't be serialized  
```
<soap:Body>
	<GetPosts>
		<count >50</count >
		<offset>100</offset>
	</GetPosts>
</soap:Body>
```

A parameters order doesn’t matter in .net framework.
We need the opportunity to replicate the server behavior using .net core because we’ve already had several client applications place parameters in arbitrary order. Could you add an ignoring of parameters order when serializing the request?

I found a [comment in the repository](https://github.com/DigDes/SoapCore/blob/develop/src/SoapCore/SoapEndpointMiddleware.cs) related with this issue:

```
// if any ordering issues, possible to rewrite like:
/*while (!xmlReader.EOF)
{
	var parameterInfo = operation.InParameters.FirstOrDefault(p => p.Name == xmlReader.LocalName && p.Namespace == xmlReader.NamespaceURI);
	if (parameterInfo == null)
	{
		xmlReader.Skip();
		continue;
	}
	var parameterName = parameterInfo.Name;
	var parameterNs = parameterInfo.Namespace;
	...
}*/
```

mtpultz:
Don't know if this solves your issue, but on the annotations there is an order param eg. `[DataMember(Order = 0)]`

https://github.com/DigDes/SoapCore/blob/4c76296d6a54bf475898ea2948422dc3df79c0da/src/SoapCore.Tests/Serialization/Models.Xml/ComplexModel1.cs

purpurproger:
Unfortunately, annotations don't solve our issue

PR #562, opened 2020-12-23
Accept any order of request arguments

purpurproger:
This PR adds the ability for clients to put request argument in different order. Now an order of request arguments needn't match the order of parameters defined in service.
Issue #543


ansongoldade:
What is the expected timing of an updated NuGet package that includes this change?

kotovaleksandr:
@ansongoldade https://www.nuget.org/packages/SoapCore/1.1.0.4
Thanks! Happy new year!

ansongoldade:
Thank you! That will be a definite contributor to making this a Happy New Year!

PR #564, opened 2021-01-05
Write error response message when encountering an invalid schema

nc-fal:
If the schema format is not correct, an exception is thrown, instead of a proper soap response, this is fixed by creating an exception and using `WriteErrorResponseMessage`


nc-fal:
@kotovaleksandr I would appreciate it if this could be added to master as well? Or should I do something to enable that?

kotovaleksandr:
@nc-fal Done! Thanks for contributing!

Issue #567, opened 2021-01-08
Support for parameterless operation contracts


rinkeb:
Hi, for some of our message contracts we need to to accept soap messages without a body. As far as i can see is currently not supported. 

For example, we need to accept echo response pings via this message:
```http
POST https://localhost:5001/service.svc HTTP/1.1
Accept-Encoding: gzip,deflate
Content-Type: text/xml;charset=UTF-8
SOAPAction: "EchoResponse"
Content-Length: 139
Host: localhost:5001
Connection: Keep-Alive
User-Agent: Apache-HttpClient/4.5.5 (Java/12.0.1)

<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/">
   <soapenv:Header/>
   <soapenv:Body/>
</soapenv:Envelope>
```
Unfortunately this results in the following exceptions:

Exception when generating the wsdl on service startup:
```csharp
System.ArgumentOutOfRangeException
  HResult=0x80131502
  Message=Index was out of range. Must be non-negative and less than the size of the collection. 
  Source=System.Private.CoreLib
  StackTrace:
   at System.ThrowHelper.ThrowArgumentOutOfRange_IndexException()
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at SoapCore.Meta.MetaBodyWriter.GetMessageContractBodyType(Type type)
   at SoapCore.Meta.MetaBodyWriter.WriteParameters(XmlDictionaryWriter writer, SoapMethodParameterInfo[] parameterInfos, Boolean isMessageContract)
   at SoapCore.Meta.MetaBodyWriter.AddTypes(XmlDictionaryWriter writer)
   at SoapCore.Meta.MetaBodyWriter.OnWriteBodyContents(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.BodyWriter.WriteBodyContents(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.BodyWriterMessage.OnWriteBodyContents(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.Message.WriteBodyContents(XmlDictionaryWriter writer)
   at SoapCore.Meta.MetaMessage.OnWriteBodyContents(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.Message.OnWriteMessage(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.Message.WriteMessage(XmlDictionaryWriter writer)
[...]

PR #571, opened 2021-01-10
#567 Support for parameterless operation contracts.

rinkeb:
PR for #567 - Support for parameterless operation contracts.

Adding support for parameterless requests which currently fail, at least in soap11. Also created some raw soap11 request tests for proofing.

This PR replaces #569, which i closed due to git related issues.


kotovaleksandr:
Great! Thanks!

Issue #574, opened 2021-01-11
New feature external WSDL and XSD


jonas-jaderberg:
Hi,
I did discover some problems with service description when you have a rather complex WSDL and schemas. I haven't investigated all problems and I see from the issues that there are some related to namespaces, prefix schemas etc. I am happy to take a look at that later.
Meanwhile, I have created a work around that I am happy to share if anyone is interested.
When you have a WSDL and Schema from another source e.g. a standard that you should implement this could also be a way to use exactly that contract.
What the workaround does is that instead fo creating the service description from the code read the WSDL och possible schemas that the wsdl uses from file. The same files are used to generated code with dotnet-svcutil for the service itself. So the source is the same.
The URL's inside the WSDL e.g. for schema imports and address location is dynamically updated.
The feature should not interfere with the normal way so there is an option to use this, default is no.
I am happy to share and create a PR if there is an interest of this.

kotovaleksandr:
Hi @jonas-jaderberg Yes, current implementation of SoapCore have limits in WSDL generation. I can suggest store wsdl files as static file on web server (like nginx) ahead SOAP service and proxy any urls which contains wsdl requests.
But this way not for everyone, please send PR with tests and changed readme.md file with instructions.

jonas-jaderberg:
Ok, I will

jonas-jaderberg:
@kotovaleksandr, excellent!  no rush but in near future an updated nuGet would be wonderful

PR #576, opened 2021-01-11
Feature external wsdl and xsd

jonas-jaderberg:
Here is the new feature to have external WSDL och XSD files stored on the server as mentioned in #574


jonas-jaderberg:
Finally :-) Was a simple misspelled configuration in test service and linux is case sensitive..

Issue #575, opened 2021-01-11
Request MessageContract parameters are not made available


imqqmi:
I need to create an endpoint based on wsdl from an existing server that I want to recreate as a test server. When I make a DataContract the parameters/properties in the class work as expected, when the request is received they are available:

```
[ServiceContract]
public interface ISampleService
{
	[OperationContract]
	TestModelResponse DoGetAuditPlanRequest(Test data);
}

[DataContract]
public class Test
{
	[DataMember]
	public string referenceID { get; set; }
}

public class SampleService : ISampleService
{
	public TestModelResponse DoGetAuditPlanRequest(Test data)
	{
		Console.WriteLine("Input data. referenceID: {0}", data.ReferenceID);
		return new TestModelResponse
		{
			referenceID = data.ReferenceID
		};
	}
}

Request:
<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/">
  <s:Header>
    <Action s:mustUnderstand="1" xmlns="http://schemas.microsoft.com/ws/2005/05/addressing/none">http://tempuri.org/ISampleService/DoGetAuditPlanRequest</Action>
  </s:Header>
  <s:Body>
    <DoGetAuditPlanRequest xmlns="http://tempuri.org/">
      <data xmlns:d4p1="http://schemas.datacontract.org/2004/07/Models" xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
        <d4p1:ReferenceID>1234</d4p1:ReferenceID>
      </data>
    </DoGetAuditPlanRequest>
  </s:Body>
</s:Envelope>

Response, as expected:
<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <s:Body>
        <DoGetAuditPlanRequestResponse xmlns="http://tempuri.org/">
            <DoGetAuditPlanRequestResult xmlns:d4p1="http://schemas.datacontract.org/2004/07/Models" xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
                <d4p1:ReferenceID>1234</d4p1:ReferenceID>
            </DoGetAuditPlanRequestResult>
        </DoGetAuditPlanRequestResponse>
    </s:Body>
</s:Envelope>
```
[...]

PR #580, opened 2021-01-13
Messagecontract, wrapped and parameters

jonas-jaderberg:
#575 I think is in the same are as I was fixing earlier so tried to extend my tests to include similar case as @imqqmi mention.
I think I found issues related to handle parameters for MessageContracts and wrapped parameters this and might fix issue #575
Can someone check/verify that this looks correct?


imqqmi:
I've tried this commit but it doesn't change the outcome for me unfortunately (issue #575), the parameter is still empty.

I've cleaned and rebuild the SoapCore library but that didn't change things either.

jonas-jaderberg:
@imqqmi, hmm strange if you look at the test cases / models I addad, test Soap11MessageContract isn't that very similar to your case?

imqqmi:
It doesn't test receiving the parameters sent in the request. This is the body that is sent in the request (empty body):
```
<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/">
  <soapenv:Header/>
  <soapenv:Body/>
</soapenv:Envelope>
```

And it responds with OK, and that's what I've been able to do too, but it doesn't receive the request parameters. So a correct test would be sending a [messagecontract] with a [MessageBodyMember] with for example a string or int. Then read the string or int and send it to the response. If the response is the same as was sent, the test completes successfully.

jonas-jaderberg:
Wrong test case I think. This one in SoapCore.Tests.MessageContract does
	public async Task Soap11MessageContract()
		{
			const string body = @"<soapenv:Envelope xmlns:soapenv=""http://schemas.xmlsoap.org/soap/envelope/"" xmlns:tem=""http://tempuri.org"">
  <soapenv:Header/>
  <soapenv:Body>	  
    <tem:MessageContractRequest>	   
        <tem:ReferenceNumber>1</tem:ReferenceNumber>
    </tem:MessageContractRequest>
  </soapenv:Body>
</soapenv:Envelope>
";
			using (var host = CreateTestHost())
			using (var client = host.CreateClient())
			using (var content = new StringContent(body, Encoding.UTF8, "text/xml"))
			using (var res = host.CreateRequest("/Service.asmx").AddHeader("SOAPAction", @"""DoRequest""").And(msg => msg.Content = content).PostAsync().Result)
			{
				res.EnsureSuccessStatusCode();

				var response = await res.Content.ReadAsStringAsync();
				Assert.IsTrue(response.Contains("MessageContractResponse"));
				Assert.IsTrue(response.Contains("ReferenceNumber>1</"));
			}
		}

imqqmi:
Thanks for the heads up, didn't know how PRs worked. I assumed it was already merged into dev branch, my bad. I'll investigate further how it is done in the test so I can change my code to match it. Thanks!

imqqmi:
After looking at the Soap11MessageContract() test, is seems it doesn't pass through the deserializer, so the defined class/model isn't created and hydrated with the values in this test, it only checks the response xml content body for containing the parameter. Would be great if the test also covers deserialization.

jonas-jaderberg:
Yes, these are "raw" tests. There are other serialization test available, not sure if they cover your case

jonas-jaderberg:
Anyway, I think this can me merged into develop

kishorekodru:
is this merged to dev branch?

kotovaleksandr:
@kishorekodru yes

kishorekodru:
I just got the latest from dev branch and i still get the message where request / part is empty 
</wsdl:types>
<wsdl:message name="NPADRVBKPort_NPADRVBKOperation_InputMessage"/>
<wsdl:message name="NPADRVBKPort_NPADRVBKOperation_OutputMessage">
<wsdl:part name="parameters" element="tns:NPADRVBKOperationResponse"/>
</wsdl:message>

kishorekodru:
<wsdl:message name="NPADRVBKPort_NPADRVBKOperation_InputMessage"/> doesnt have part name/element.

kishorekodru:
@kotovaleksandr  i got the latest from dev branch. still doesnt generate the part? Any suggestions? 
            app.UseSoapEndpoint<NPADRVBKService>("/Intake/npadrvbk", new BasicHttpBinding(), SoapSerializer.XmlSerializer);

This is what I have .

jonas-jaderberg:
@kishorekodru, how does your contract/interface look like?
If there is now input parameter, isn't the part unnecessary? 
I have a wsdl with an operation without parameter and it looks like
..
<message name="pullSnapshotDataInputMessage"/>
<message name="pullSnapshotDataOutputMessage">
<part name="return" element="tns:pullSnapshotDataOutput"/>
</message>
..
and it's a valid WSDL and complies to WS-I.

kishorekodru:
Interface  : 
    [ServiceContract]
    interface NPADRVRLInterface
    {
        [OperationContract]
        public NPADRVRLData NPADRVRLOperation(NPADRVRLData input);
    }

Impl: 
    public class NPADRVRLService : NPADRVRLInterface
    {
        public NPADRVRLData NPADRVRLOperation(NPADRVRLData input)
        {
            IntakeSOAPService svc = new IntakeSOAPService();
            return (NPADRVRLData)svc.ProcessProgram(input, "NPADRVRL");
        }
    }

I am new to .NET Core and it has been annoying with only response being generated and not input in the wsdl

jonas-jaderberg:
Ok, I misunderstood thought you had no input parameter. But, you have. How does your NPADRVRLData look like, is it a MessageContract?

kishorekodru:
It is a dataContract. Should i make it MessageContract?

jonas-jaderberg:
If you like use SoapSerializer.XmlSerializer MessageContract gives more control. Datacontracts is used with DataContractSerializer but not sure what happens if you mix

PR #589, opened 2021-02-10
fix XmlElementAttribute array serialization

cubidobusinesssolutions:
#243


cubidobusinesssolutions:
No need to fix anything at all :) I will post a solution to #243

PR #605, opened 2021-03-18
Fix issue with container-less arrays on MessageContracts

rickykaare:
This PR solves an issue with serializing arrays without a container element on MessageContracts that contains MessageHeader properties. 

The issue was discovered when trying to use a contract generated from wsdl with the `dotnet-svcutil`. The generated code uses MessageContracts with MessageHeaders, which triggers special logic in the SoapCore XmlSerializer.

An example of a failing contract could look like this (A similar example has been added to the unit tests):
```csharp
[MessageContract]
public class Contract
{
    [MessageHeader]
    public ComplexModel1 Header { get; set; }

    [MessageBodyMember]
    [XmlElement("Array")]
    public ComplexModel1[] Array { get; set; }

    [MessageBodyMember]
    public ComplexModel SomeOtherMember { get; set; }
}
```

Which renders the Array elements side-by-side without a container element, like this:
```xml
<Contract>
    <Array>...</Array> 
    <Array>...</Array>
    <Array>...</Array>
    <SomeOtherMember>...</SomeOtherMember>
</Contract>
```

The fix includes:
* Passing the correct `MemberInfo` to `DeserializeInputParameter()`
* Prevent reading container element if array has the `XmlElementAttribute` which indicate that there is no container element.
* Use the element name from `XmlElementAttribute` if present.
* Add logic for returning empty arrays instead of null.

### DIFF
--- a/src/SoapCore.Tests/Serialization/Models.Xml/ISampleService.cs
+++ b/src/SoapCore.Tests/Serialization/Models.Xml/ISampleService.cs
@@ -76,6 +76,10 @@ namespace SoapCore.Tests.Serialization.Models.Xml
 		[XmlSerializerFormat(SupportFaults = true)]
 		UnwrappedStringMessageBodyMemberResponse TestUnwrappedStringMessageBodyMember(BasicMessageContractPayload x);
 
+		[OperationContract(Action = ServiceNamespace.Value + nameof(TestMessageContractWithArrays), ReplyAction = "*")]
+		[XmlSerializerFormat(SupportFaults = true)]
+		MessageContractResponseWithArrays TestMessageContractWithArrays(MessageContractRequestWithArrays request);
+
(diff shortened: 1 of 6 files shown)


EamonHetherton:
This is breaking when the request paramters contains an empty array element.

e.g. <someArrayParameter />

EamonHetherton:
have created a new pull request to show problem as #622 as I couldn't commit to this pull request directly.

Mazafakaa (commented):
> This is breaking when the request parameters contains an empty array element.
> 
> e.g.
> `<someArrayParameter />`

I noticed this problem when I passed in an array with the specified namespace, elements without specifying the namespace.

e.g.
`<ns:SomeArray>
<SomeArrayElement></SomeArrayElement>
<SomeArrayElement></SomeArrayElement>
<SomeArrayElement></SomeArrayElement>
</ns:SomeArray>`

EamonHetherton:
Can you add a failing test to demonstrate?

PR #607, opened 2021-03-24
Fix a couple of bugs found when importing existing WSDL

brianfeucht:
I am working on integrating this library with an existing [WSDL](https://soapapi.litmusapp.com/2010-06-21/api.asmx?WSDL) and ran into a couple of bugs.

First this `UseSoapEndpoint` extension was not passing the `WsdlFileOptions` to the wrapped function.

The second bug I ran into was a case where the existing WSDL didn't have a `schemaLocation` which triggered a null reference exception.  I handled this case by adding the attribute, but I'd also be open to doing nothing if the attribute is missing.


brianfeucht:
After thinking about this a bit more, I'm leaning towards doing nothing in the null case.  Please let me know if you have opinions either way.

Issue #613, opened 2021-04-01
XmlDictionaryWriter.CreateDictionaryWriter(xmlTextWriter) in SoapMessageEncoder.cs changed behaviour in commit  472eb5e


iBrotNano:
The bug is introduced in commit 472eb5ef.

Before the `XmlDictionaryWriter` was created with `XmlDictionaryWriter.CreateTextWriter(stream, _writeEncoding, false);`. After the change it ist created with `XmlDictionaryWriter.CreateDictionaryWriter(xmlTextWriter);`. This breaks the communication to a third party application for me.

By stepping through the .NET Code i realized that the two factory methods are returning different implementations of a XmlDictionaryWriter.

The new method returns this:

```C#
writer = new XmlUtf8RawTextWriterIndent(output, this);
```

The old method returns:

```C#
XmlUTF8TextWriter writer = new XmlUTF8TextWriter();
writer.SetOutput(stream, encoding, ownsStream);
var asyncWriter = new XmlDictionaryAsyncCheckWriter(writer);
return asyncWriter;
```

Both are wrapped in a XmlWrappedWriter.

PS: I figured out that it is caused by the XML indentation.

iBrotNano:
A resolution is merged. I close this issue.

PR #615, opened 2021-04-07
#613 Added XML indentation and omitting XML declaration at the UseSoapEndpoint extension methods.

iBrotNano:


iBrotNano:
I am not shure how to write tests for the commit. There are no tests right now for the methods at all. Because they are extensions methods which are static it is not so easy to test them.

Issue #620, opened 2021-04-14
ServiceContract namespace is used when constructing an element in MessageHeader


pxk911:
Hello, I am having an issue that when constructing an element in the header of a Soap message, the namespace defined in the [ServiceContract] is used instead of the namespace defined in [MessageHeader].

For example, the correct message I want to achieve is :
```csharp
﻿<s:Envelope >
  <s:Header>
    <Header_Element version="Test header" xmlns="http://header_namespace.org" />
  </s:Header>
  <s:Body>
    <Response_Element version="Test header as body" xmlns="http://response_namespace.org" />
  </s:Body>
</s:Envelope>
```
And achieveing the correct message with the following definitions,

My MessageContract definition 
```csharp
[XmlSerializerFormat]
[MessageContract(IsWrapped = false)]
public class A_Simple_Response
{
    [MessageHeader(Name = "Header_Element",Namespace ="http://header_namespace.org")] //namespace for header element is defined here
    public A_Simple_Response_Header Response_Header;

    [MessageBodyMember(Name = "Response_Element", Namespace = "http://response_namespace.org")]
    public A_Simple_Response_Header Response_Header_As_Body;

}
```

My ServiceContract definition:
```csharp
[ServiceContract] //ServiceContract namespace should not be set
[XmlSerializerFormat]
interface IXmlTestService
{
    [OperationContract]
    public A_Simple_Response Simple_Ping();
}
```

What I get is the element in the header is using a namespace"http://tempuri.org/" which I believe is because I have not set the namespace in the [ServiceContract] but I want the namespace to be the one I defined in the [MessageHeader].

```csharp
<s:Envelope >
  <s:Header>
    <Header_Element version="Test header" xmlns="http://tempuri.org/" /> //wrong namespace
  </s:Header>
  <s:Body>
    <Response_Element version="Test header as body" xmlns="http://response_namespace.org" /> //namespace in the body is correct
  </s:Body>
</s:Envelope>
```
I found in the SoapEndpointMiddleware.cs at line 445, it seems that it's using the operation.Contract.Namespace(the namespace defined in the [ServiceContract]) instead of the namespace defined in the [MessageHeader].

https://github.com/DigDes/SoapCore/blob/b87e2547386e7dd3e30075f20814a548e3f06352/src/SoapCore/SoapEndpointMiddleware.cs#L442-L446

Is this a bug or is it designed to be this way?
if it's a feature, how to achieve the correct message?

PR #621, opened 2021-04-14
Fix for issue #620 ServiceContract namespace is used when constructing an element in MessageHeader

pxk911:
A fix for issue #620


kotovaleksandr:
@pxk911 Hi! Thanks for contributing! Please, don't change exists test which cover real cases. Add new test for your issue.

kotovaleksandr:
Thanks!

Issue #625, opened 2021-04-20
byte[] is not deserialized by SoapSerializer.XmlSerializer


IvanLeonenko:
I'm using `SoapSerializer.XmlSerializer` and trying to get simple web method to work (see below).

Basically, I want to pass byte array. I see that in generated wsdl type for the parameter is correct - `s:base64Binary`.

But when I try to send request (inline base64 payload dGVzdCBzdHJpbmc=) from generated client code or from SoapUI I get the following error:
```
    <s:Fault>
      <faultcode>s:Client</faultcode>
      <faultstring>End element 'data' from namespace 'http://test/testsvc' expected. Found text 'dGVzdCBzdHJpbmc='. Line 1, position 231.</faultstring>
    </s:Fault>
```

Same request works fine in old .NET service, no erros.

Request
```
<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:tes="http://test/testsvc">
   <soapenv:Header/>
   <soapenv:Body>
      <tes:Test>
         <tes:data>dGVzdCBzdHJpbmc=</tes:data>
      </tes:Test>
   </soapenv:Body>
</soapenv:Envelope>
```

Service
```
[ServiceContract(Namespace = "http://test/testsvc")]
    public interface ITestService
    {
        [OperationContract]
        ResultResponse Test(byte[] data);
    }

    public class ResultResponse
    {
        public bool Result { get; set; }
        public string Base64 { get; set; }
    }

    public class TestService : ITestService
    {
        public ResultResponse Test(byte[] data)
        {
            return new ResultResponse() {Result = true, Base64 = Convert.ToBase64String(data)};
        }
    }
```

I noticed that in `SerializerHelper` there's [a check](https://github.com/DigDes/SoapCore/blob/491bd57c138eb6074ad54433a8fb652dc92c53b9/src/SoapCore/SerializerHelper.cs#L38)  after which Byte[] is handled in `DeserializeArrayXmlSerializer`  and generates the error. If I call `DeserializeObject` for Byte[] type instead of `DeserializeArrayXmlSerializer` then parameter is deserialized correctly.

Is it a bug, or I'm doing something wrong?

Thank you

IvanLeonenko:
#626

kotovaleksandr:
fixed in #626

samuel-barbosa:
I could see that the fix was really effective for methods with _single dimension_ array of byte arguments.

[...]

PR #626, opened 2021-04-20
Fix for #625 Support deserialization from base64 to byte[] in input parameters. Ad…

IvanLeonenko:
- Support deserialization from base64 to byte[] in input parameters
- Byte array unit test
- Fix XmlArray and XmlElement attribute reading for input parameters, before method attributes were used instead of parameter attributes. Depend on ICustomAttributeProvider instead of MemberInfo in DeserializeArrayXmlSerializer. (Introduced in https://github.com/DigDes/SoapCore/commit/54bc556f96a94218e873ee4b44a8425fc029f573#diff-8bb22c90c9d99146addfac0efb111b8fa0a045dd97ce8b670e18185d9e520340) If you debug you can see that MemberInfo is from method not a parameter, and we want to get parameter attributes as I understand.


IvanLeonenko:
Fixes #625

PR #628, opened 2021-04-23
Simplified use of messageHeaderAttribute.Namespace. Added support for messageHeaderAttribute.MustUnderstand.

intorr:


kotovaleksandr:
@intorr Thanks for contributing! Please add test for new behavior.

PR #681, opened 2021-07-17
Optional parameter named "Body" and support for array parameters

preeder:
Fixed bug where an optional parameter which is not included with the same local name as a SOAP element would cause deserialization to enter an endless loop

Added support for array parameters


kotovaleksandr:
@preeder Please add test which cover bug and fix exists tests - CI is broken

andersjonsson:
Closing this PR. Please resolve conflicts and re-open if still relevant

PR #683, opened 2021-07-19
cleanup TrailingServicePathTuner

tibel:
Cleanup TrailingServicePathTuner a bit.

It is useful without endpoint routing only, I guess.
Could `app.MapWhen()` or `endpoints.Map()` be used alternatively?


kotovaleksandr:
> Could app.MapWhen() or endpoints.Map() be used alternatively?
I think is not necessary, few people use it

PR #699, opened 2021-08-10
Fix for issue 414

ThisBytes5:
Test create this time, and PR into develop instead of master.

### DIFF
--- a/src/SoapCore/HeadersHelper.cs
+++ b/src/SoapCore/HeadersHelper.cs
@@ -2,6 +2,8 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using System.ServiceModel.Channels;
+using System.Xml;
 using Microsoft.AspNetCore.Http;
 
 namespace SoapCore
@@ -10,9 +12,20 @@ namespace SoapCore
 	{
 		private static readonly char[] ContentTypeSeparators = new[] { ';' };
 
+		
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
-		public static string GetSoapAction(HttpContext httpContext, System.Xml.XmlDictionaryReader reader)
+		public static string GetSoapAction(HttpContext httpContext, ref Message message)
 		{
+			XmlDictionaryReader reader = null;
+			if (!message.IsEmpty)
+			{
+				MessageBuffer mb = message.CreateBufferedCopy(int.MaxValue);
+				Message responseMsg = mb.CreateMessage();
+				message = mb.CreateMessage();
+
+				reader = responseMsg.GetReaderAtBodyContents();
+			}
+
 			var soapAction = httpContext.Request.Headers["SOAPAction"].FirstOrDefault();
 			if (soapAction == "\"\"")
 			{
--- a/src/SoapCore/SoapEndpointMiddleware.cs
+++ b/src/SoapCore/SoapEndpointMiddleware.cs
@@ -206,6 +206,9 @@ namespace SoapCore
 				return;
 			}
 
+			var soapAction = HeadersHelper.GetSoapAction(httpContext, ref requestMessage);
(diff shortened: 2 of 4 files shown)


src/SoapCore/HeadersHelper.cs
```diff
@@ -10,9 +12,20 @@ internal static class HeadersHelper
 	{
 		private static readonly char[] ContentTypeSeparators = new[] { ';' };
 
+		
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
-		public static string GetSoapAction(HttpContext httpContext, System.Xml.XmlDictionaryReader reader)
+		public static string GetSoapAction(HttpContext httpContext, ref Message message)
```
tibel, line 17: This shouldn't be changed.

src/SoapCore/SoapEndpointMiddleware.cs
```diff
@@ -206,6 +206,9 @@ private async Task ProcessOperation(HttpContext httpContext, IServiceProvider se
 				return;
 			}
 
+			var soapAction = HeadersHelper.GetSoapAction(httpContext, ref requestMessage);
+			requestMessage.Headers.Action = soapAction;
```
tibel, line 210: Just move the lines up, so you can use the reader without changes. 

https://github.com/DigDes/SoapCore/blob/37d3e2da6933a544ea9e7a2bd74201ad55bc8405/src/SoapCore/SoapEndpointMiddleware.cs#L225-L231

ThisBytes5: Doing so puts the message in a read state, which in turn causes an exception. hence why I changed the GetSoapAction. The message needs to be copied, before reading it. I thought about overloading GetSoapAction, but it's only used in the one place. Suggestions?

ThisBytes5: no test is failing now, when I simply moved the lines around and ran the UT's (didn't see them the first time around as I had the sample solution open) the message inspector tests failed.

Yes, it's in a read state, and it can't be copied once it's state is no longer created. So the only fix I found was to move the getSoapAction before the call for IMessageInspector2 and modify the GetSoapAction, and to copy the message before reading the action.

Yes it may copy the message when there isn't a IMessageInspector assigned, but the IMessageInspector and GetSoapAction don't know about each other, and I didn't want to change that, so the copy in the method was the only thing I came up with.

tibel: Just checked that `MessageInspectorMock` is creating a copy of the message and this is not allowed after creating `message.GetReaderAtBodyContents()`. That's why the tests are failing.

So we have two options now:
- copy the message before (as you did)
- pass the `XmlDictionaryReader` to `IMessageInspector` (would be breaking but better for performance)

Another question is:
What should be possible/allowed in a message inspector?

ThisBytes5: Looking at our code for WCF (which I'm replacing with SoapCore) the signatures are as follows:

public object AfterReceiveRequest(ref Message request, IClientChannel channel, InstanceContext instanceContext)
public void BeforeSendReply(ref Message reply, object correlationState)

Which to me says the following:
1) copy message before, as I did
2) anything can be done in the message inspector since you can replace the message and in fact do if you attempt to read it at all.

So that being said I wouldn't change the code as it's written. However, maybe consider giving another message inspector that passes the XmlDictionaryReader for those times where people only need to read the message and not modify it (which currently is my case).

ThisBytes5: Not to be a pain, but where do you stand on a decision for this? I need to report on my project progress and this is sort of holding me up at the moment. Worst case I pull the code directly into my project for the time being.

ThisBytes5: Any update on this? Are you waiting for something from me?

PR #725, opened 2021-10-18
Improve MessageEncoder matching

### LINKED ISSUE #724: MessageEncoder fails to find a match if Request.ContentType doesn't contain a charset
As I am porting an old web service api https://api.legaonline.se/rentalApi.asmx I had some trouble with a collection of postman scripts. My scripts uses soap12 and the new service only works if I add the Soap12 MessageEncoder first.

Turns out that the ContentType of my postman requests was `application/soap+xml`. In that case it doesn't match a MessageEncoder with Soap12 and UTF8. That led to the server picking the first MessageEncoder as a fallback, and if that was Soap11 I got an error.

I'll submit a PR that allows Request.ContentType to be empty and that does a second pass to find a MessageEncoder with the correct MediaType as fallback, so that you will at least get the right protocol version.

andersjonsson:
- Allow empty request charset
- Add a fallback pass without charset comparison at all

fixes #724

### DIFF
--- a/src/SoapCore/MessageEncoder/SoapMessageEncoder.cs
+++ b/src/SoapCore/MessageEncoder/SoapMessageEncoder.cs
@@ -63,14 +63,14 @@ namespace SoapCore.MessageEncoder
 
 		public XmlDictionaryReaderQuotas ReaderQuotas { get; }
 
-		public bool IsContentTypeSupported(string contentType)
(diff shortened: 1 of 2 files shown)


excalibur1028:
Hello, when will this commit be released on NuGet? Thanks

PR #730, opened 2021-10-19
made sure that outparameters are added to the return type of the operation

### LINKED ISSUE #729: Trouble with "Add Service Reference"
When trying to add a service reference to my ported Api i found a few weird issues.

image

When I click on the "Go"-button, four requests are fired off.

GET /RentalApi.asmx/$metadata
GET /RentalApi.asmx/mex

POST /RentalApi.asmx
POST /RentalApi.asmx/mex

The GET for $metadata gets a 404 and the POSTS crash. Null Reference Exception.

image

Not sure if the POSTS occur because the GET for /$metadata fails or what's up.

Do you have any idea what the correct behavior would be? I can do some more digging tomorrow, but I wanted to check with you first

andersjonsson:
Fixes #729
All unit tests still work and I've added one for this case too

### DIFF
--- a/src/SoapCore/Meta/MetaBodyWriter.cs
+++ b/src/SoapCore/Meta/MetaBodyWriter.cs
@@ -248,19 +248,7 @@ namespace SoapCore.Meta
 						hasWrittenSchema = true;
 					}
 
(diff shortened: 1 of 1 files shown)


andersjonsson:
Thank you for the speedy merge! Would it be possible to get a new nuget version later today?

PR #744, opened 2021-10-28
add LockingXmlNamespaceManager

### LINKED ISSUE #743: XmlNamespaceManager corruption
When testing things on our servers I got this exception:
```
System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
   at System.Collections.Generic.Dictionary<TKey,TValue>.FindValue(TKey key)
   at System.Collections.Generic.Dictionary<TKey,TValue>.TryGetValue(TKey key, TValue& value)
   at System.Xml.XmlNamespaceManager.LookupNamespaceDecl(String prefix)
   at System.Xml.XmlNamespaceManager.LookupNamespace(String prefix)
   at System.Xml.XmlNamespaceManager.LookupPrefix(String uri)
   at SoapCore.Namespaces.AddNamespaceIfNotAlreadyPresentAndGetPrefix(XmlNamespaceManager xmlNamespaceManager, String preferredPrefix, String uri)
   at SoapCore.Namespaces.AddDefaultNamespaces(XmlNamespaceManager xmlNamespaceManager)
   at SoapCore.SoapEndpointMiddleware<T>.GetXmlNamespaceManager()
   at SoapCore.SoapEndpointMiddleware<T>.ProcessMeta(HttpContext httpContext)
   at SoapCore.SoapEndpointMiddleware<T>.Invoke(HttpContext httpContext, IServiceProvider serviceProvider)
   at Microsoft.AspNetCore.Routing.EndpointMiddleware.<Invoke>g__AwaitRequestTask|6_0(Endpoint endpoint, Task requestTask, ILogger logger)
   at legaonline.apiws.Startup.<>c.<<Configure>b__4_0>d.MoveNext() in Startup.cs:line 112
--- End of stack trace from previous location ---
   at StackExchange.Exceptional.ExceptionalMiddleware.Invoke(HttpContext context)
```

I see that this kind of error has been encountered before and that a PR has been merged #693 but obviously it can still happen, if multiple people access the endpoint before the XmlNamespaceManager has been fully populated

andersjonsson:
Fixed #743 

I'm not fond of this solution but it should fix these issues once and for all. I think it would be possible to restructure the code to fix this in another way, but that is much more invasive.

This is a possibly breaking change, but it should be easy to fix.
What do you think?

### DIFF
--- a/src/SoapCore.Tests/FaultExceptionTransformer/TestFaultExceptionTransformer.cs
+++ b/src/SoapCore.Tests/FaultExceptionTransformer/TestFaultExceptionTransformer.cs
@@ -9,7 +9,7 @@ namespace SoapCore.Tests.FaultExceptionTransformer
 {
 	public class TestFaultExceptionTransformer : IFaultExceptionTransformer
 	{
-		public Message ProvideFault(Exception exception, MessageVersion messageVersion, Message requestMessage, XmlNamespaceManager xmlNamespaceManager)
+		public Message ProvideFault(Exception exception, MessageVersion messageVersion, Message requestMessage, LockingXmlNamespaceManager xmlNamespaceManager)
 		{
 			var fault = new TestFault
 			{
--- a/src/SoapCore.Tests/MessageContract/RawRequestSoap11Tests.cs
+++ b/src/SoapCore.Tests/MessageContract/RawRequestSoap11Tests.cs
@@ -139,7 +139,7 @@ namespace SoapCore.Tests.MessageContract
 				var root = new XmlDocument();
 				root.LoadXml(response);
 
-				var nsmgr = new XmlNamespaceManager(root.NameTable);
+				var nsmgr = new LockingXmlNamespaceManager(root.NameTable);
 				nsmgr.AddNamespace("wsdl", "http://schemas.xmlsoap.org/wsdl/");
 				nsmgr.AddNamespace("xsd", "http://www.w3.org/2001/XMLSchema");
 
@@ -176,7 +176,7 @@ namespace SoapCore.Tests.MessageContract
 				var root = new XmlDocument();
 				root.LoadXml(response);
 
-				var nsmgr = new XmlNamespaceManager(root.NameTable);
+				var nsmgr = new LockingXmlNamespaceManager(root.NameTable);
 				nsmgr.AddNamespace("wsdl", "http://schemas.xmlsoap.org/wsdl/");
 				nsmgr.AddNamespace("xsd", "http://www.w3.org/2001/XMLSchema");
 
@@ -213,7 +213,7 @@ namespace SoapCore.Tests.MessageContract
 				var root = new XmlDocument();
 				root.LoadXml(response);
 
-				var nsmgr = new XmlNamespaceManager(root.NameTable);
+				var nsmgr = new LockingXmlNamespaceManager(root.NameTable);
 				nsmgr.AddNamespace("wsdl", "http://schemas.xmlsoap.org/wsdl/");
 				nsmgr.AddNamespace("xsd", "http://www.w3.org/2001/XMLSchema");
 
--- a/src/SoapCore.Tests/Wsdl/WsdlTests.cs
+++ b/src/SoapCore.Tests/Wsdl/WsdlTests.cs
@@ -257,7 +257,7 @@ namespace SoapCore.Tests.Wsdl
 			var root = new XmlDocument();
 			root.LoadXml(wsdl);
 
-			var nsmgr = new XmlNamespaceManager(root.NameTable);
+			var nsmgr = new LockingXmlNamespaceManager(root.NameTable);
 			nsmgr.AddNamespace("wsdl", "http://schemas.xmlsoap.org/wsdl/");
 			nsmgr.AddNamespace("xs", "http://www.w3.org/2001/XMLSchema");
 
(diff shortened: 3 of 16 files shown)


TsengSR:
@andersjonsson 
I wonder why you have replaced `XmlNamespaceManager` with `LockingXmlNamespaceManager` in every single class. Since later inherits from first, it should have been enough to add the `LockingXmlNamespaceManager` and only replaces the positions where you `new` it. Half of the other changes seems completely unnecessarily.

But even that seems unnecessarily, to only add a new class because of locking. Putting `lock` in that many places will **seriously** hurt the performance.

Making `DefaultFaultExceptionTransformer` scoped should have (see [here](https://github.com/DigDes/SoapCore/blob/develop/src/SoapCore/SoapEndpointExtensions.cs#L345)) solved the issue in the first place because then the `XmlNamespaceManager` would then be scoped per request and each user calling the endpoint would have got his own instance of it and no concurrency issue appear in the first place.

andersjonsson:
> @andersjonsson I wonder why you have replaced `XmlNamespaceManager` with `LockingXmlNamespaceManager` in every single class. Since later inherits from first, it should have been enough to add the `LockingXmlNamespaceManager` and only replaces the positions where you `new` it. Half of the other changes seems completely unnecessarily.

I replaced it everywhere just to make sure that you couldn't end up with a regular XmlNamespaceManager.

> Putting `lock` in that many places will **seriously** hurt the performance.

Agreed. I don't like that solution either. And to be honest I wasn't expecting it to be merged without a discussion 😉 

>Making DefaultFaultExceptionTransformer scoped should have (see here) solved the issue in the first place because then the XmlNamespaceManager would then be scoped per request and each user calling the endpoint would have got his own instance of it and no concurrency issue appear in the first place.

I might be missing something, but I don't see how `DefaultFaultExceptionTransformer` is involved in this? See my stack trace in #743 


Now, after looking at it again I think this issue is because of https://github.com/DigDes/SoapCore/blob/337a298a2e24c65fd4b9b200c154baf7975bf27b/src/SoapCore/SoapEndpointMiddleware.cs#L810

Since middlewares are, to my understanding, singletons, each call gets the same XmlNamespaceManager (if you use XmlNamespacePrefixOverrides).

If I'm correct, the most straightforward solution would be to copy the namespaces in XmlNamespacePrefixOverrides to a new XmlNamespaceManager instance and return the new instance.

If that's the case I would be happy to send a new PR where I remove the LockingXmlNamespaceManager again.

What do you think?

TsengSR:
> > Making DefaultFaultExceptionTransformer scoped should have (see here) solved the issue in the first place because then the XmlNamespaceManager would then be scoped per request and each user calling the endpoint would have got his own instance of it and no concurrency issue appear in the first place.
  > 
  > I might be missing something, but I don't see how `DefaultFaultExceptionTransformer` is involved in this? See my stack trace in #743
  
It's one of the places where `IFaultExceptionTransformer` is resolved and where `XmlNamespaceManager` is instantiated and it's a singleton which means it's shared between all requests
  

https://github.com/DigDes/SoapCore/blob/337a298a2e24c65fd4b9b200c154baf7975bf27b/src/SoapCore/SoapEndpointMiddleware.cs#L675-L687

While not directly related to the executed path, it's still subject of race conditions that two separate requests may access the same path at the same time, though maybe less likely than the one you stumbled uppon.
  
  > Now, after looking at it again I think this issue is because of
  > 
  > https://github.com/DigDes/SoapCore/blob/337a298a2e24c65fd4b9b200c154baf7975bf27b/src/SoapCore/SoapEndpointMiddleware.cs#L810
  > 
  > Since middlewares are, to my understanding, singletons, each call gets the same XmlNamespaceManager (if you use XmlNamespacePrefixOverrides).
  
  Yea, middlewares are singletons, but they can resolve scoped or transient dependencies (either via injecting into `Invoke(HttpContext context, MyDependency dependency)` or via `context.RequestServices.Get(Required)Service<T>();`
  
  
  > If I'm correct, the most straightforward solution would be to copy the namespaces in XmlNamespacePrefixOverrides to a new XmlNamespaceManager instance and return the new instance.
  
  That's one thing yes, but the quasi-singleton instances of `XmlNamespaceManager` is the other: They are injected or instantiated in many singleton services which makes them singletons too and hence shared between requests. To solve that all registered services which `new` or inject `XmlNamespaceManager` would have to become scoped too.

> If that's the case I would be happy to send a new PR where I remove the LockingXmlNamespaceManager again.
> 
> What do you think?

andersjonsson:
> It's one of the places where `IFaultExceptionTransformer` is resolved and where `XmlNamespaceManager` is instantiated and it's a singleton which means it's shared between all requests
> 
> https://github.com/DigDes/SoapCore/blob/337a298a2e24c65fd4b9b200c154baf7975bf27b/src/SoapCore/SoapEndpointMiddleware.cs#L675-L687
> 
> While not directly related to the executed path, it's still subject of race conditions that two separate requests may access the same path at the same time, though maybe less likely than the one you stumbled uppon.

Ah I see what you mean! Wouldn't that only be an issue if it was stored in a field/property within `IFaultExceptionTransformer`?
As far as I can tell the `XmlNamespaceManager` in your example is resolved with `GetXmlNamespaceManager()` and used as a method parameter. 
If I fix `GetXmlNamespaceManager()` as I suggested we should be alright. Or am I missing something?

PR #749, opened 2021-11-01
Reengineer fix for xmlnamespacemanager corruption

andersjonsson:
@TsengSR @kotovaleksandr 

PR based on the discussion in #744
As far as I can tell `GetXmlNamespaceManager()` in `SoapEndpointMiddleware` is called everywhere that an `XmlNamespaceManager` is needed. That instance doesn't seem to be stored in any singletons.

It's unfortunate to revert this change between nuget versions, but I think it's worth it.

What do you think?

This change will be tested in our staging environment tomorrow morning, so we'll know for sure that it fixes the original issue.


andersjonsson:
Yep, this fix seems to do the trick also and I feel much better about it.
If we want pure performance it should be fairly easy to cache the namespaces per endpoint later.

andersjonsson:
I came across another potential bug, but I filed a separate issue for that

andersjonsson:
Had to convert it back to draft. I realize we need to do some more testing on our end.

andersjonsson:
Testing is complete. Works as intended

kotovaleksandr:
@TsengSR What you say? Looks good?
I'm sorry, it's my bad that i didn't look well enough at the previous PR. i should have seen performance problem with lock:(

TsengSR:
> @TsengSR What you say? Looks good? I'm sorry, it's my bad that i didn't look well enough at the previous PR. i should have seen performance problem with lock:(

Yea, looks fine

PR #756, opened 2021-11-10
added support (and test) for arrayofarrayof

andersjonsson:
We have a method that returns `List<List<Thingy>>`. That resulted in a return type of ArrayOfList'1 in the wsdl.

This PR fixes so that nested arrays/ienumerables works.

If you merge this PR I'd appreciate a new nuget package


kotovaleksandr:
Thanks! https://www.nuget.org/packages/SoapCore/1.1.0.21

PR #761, opened 2021-11-19
Message inspector 3

moinessim:
Problem:
I need a way to 
1. Validate some http header 
2. Validate Message against XmlSchema

Before deserializing request Message into typed input
And respond with a specific SOAP Message (no faults) if either fail.

Tried using MessageInspector2 and Tuner but they don't provide a seamless way to skip the pipeline and issue a response.

This PR adds a way to do just that with the IMessageInspector3.

Please let me know if this can be tackled with the existing functionality, or any other suggestion.

To do:
Add some more tests and docs.

### DIFF
--- a/src/SoapCore.Tests/MessageInspectors/InspectorStyle.cs
+++ b/src/SoapCore.Tests/MessageInspectors/InspectorStyle.cs
@@ -3,6 +3,7 @@ namespace SoapCore.Tests.MessageInspectors
 	public enum InspectorStyle
 	{
 		MessageInspector,
-		MessageInspector2
+		MessageInspector2,
+		MessageInspector3
 	}
 }
--- a/src/SoapCore.Tests/MessageInspectors/MessageInspector3/MessageInspector3Mock.cs
+++ b/src/SoapCore.Tests/MessageInspectors/MessageInspector3/MessageInspector3Mock.cs
@@ -24,10 +24,7 @@ namespace SoapCore.Tests.MessageInspectors.MessageInspector3
 			LastReceivedMessage = message;
 			AfterReceivedRequestCalled = true;
 
-			// validate message
-			ValidateMessage(ref message);
-
-			return null;
+			return ValidateMessage(ref message);
 		}
 
 		public void BeforeSendReply(ref Message reply, ServiceDescription serviceDescription, object correlationState)
@@ -35,9 +32,9 @@ namespace SoapCore.Tests.MessageInspectors.MessageInspector3
 			BeforeSendReplyCalled = true;
 		}
 
-		private void ValidateMessage(ref Message message)
+		private string ValidateMessage(ref Message message)
 		{
-			throw new FaultException(new FaultReason("Message is invalid."), new FaultCode("Sender"), null);
+			return "Failed";
 		}
 	}
 }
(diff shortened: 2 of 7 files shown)


kotovaleksandr:
Hi @moinessim! I understand problem what you want solve, but PR contains very specific implementation for one specific case. @tibel Can you help? I can suggest break IMessageInspector interface and add HttpContext as argument, but it's not resolve all @moinessim requirements.

andersjonsson:
This would probably be a useful feature anyways, but the quickest solution for your issue would be to do the work in a separate middleware, before even hitting SoapCore
I do some inspection, xml fixes and header manipulation of my own, in a middleware.

My middleware, from Startup.cs, as inspiration
```
const string Soap12NS = "http://www.w3.org/2003/05/soap-envelope";

app.Use(async (context, next) =>
{
	if (context.Request.Method == "POST")
	{
		//This line allows us to set the reader for the request back at the beginning of its stream.
		context.Request.EnableBuffering();

		//We convert the byte[] into a string using UTF8 encoding...
		var bodyAsText = Encoding.UTF8.GetString(await context.Request.Body.ReadFullyAsync());
					
		try
		{
			var xdoc = XDocument.Parse(bodyAsText);
						
			bodyAsText = xdoc.ToString();
			
			//The parsing fixes some issues with malformed xml, from old clients
			context.Request.Body = new MemoryStream(Encoding.UTF8.GetBytes(bodyAsText));

			//Logging!
			SoapMessageLog.WriteLog(context, bodyAsText);

			//Upgrade Content-Type if Envelope is Soap1.2
			if (context.Request.Headers["Content-Type"] != "application/soap+xml")
			{
				var envelope = xdoc.Descendants().FirstOrDefault(x => x.Name.LocalName == "Envelope");
				if (envelope != null && envelope.Name.ToString().Contains(Soap12NS))
				{
					context.Request.Headers["Content-Type"] = "application/soap+xml";
				}
			}

		}
		catch
		{ }

		//..and finally, assign the read body back to the request body, which is allowed because of EnableRewind()
		context.Request.Body.Position = 0;
	}

	await next();
}

);
```

This approach could be used to solve your issue also. Just do your validation and only call next() if successful. Otherwise just write to the response directly.

moinessim:
Thank you for your advice @andersjonsson.
I tried the middleware approach, but the pain point is creating a Message based on a typed object.

I'm using Xmlserializer. Message.CreateMessage uses DataContractSerializer, which gives me undesired output.

The bodyWriter of SoapCore is internal, so I can't use it from outside.

I think a public function that creates a Message from a type, using SoapCoreOptions, would be ideal.

What do you think?

andersjonsson:
Ah, gotcha!

I'm not involved in the design of SoapCore (I'm just a user that has spent a lot of time recently with replicating my own legacy service).

Just removing "internal" from BodyWriter would be the quickest solution, but might not be desired.

If it was up to me I'd explore breaking apart ProcessOperation in SoapEndpointMiddleware so that it acts as it's own middleware pipe.

ProcessOperation (the part that reads the requestMessage) -> (any user defined MessageProcessorMiddlewares) -> (the rest of ProcessOperation).

A MessageProcessorMiddleware could work like a normal asp.net core middleware and just "return await next();" if everything was in order, or return a custom message. BodyWriter could be supplied as a parameter.

One would have to decide if those middlewares should have access to httpContext, or if the return value would be both a Message as well as headers, statuscode etc.

That's just off the top of my head. I have no idea if that design would hold up in the real world, or if it's in conflict with other design plans for the future. @kotovaleksandr ?

This design could solve many of the same things that MessageInspector does, but allow you to short-circuit the processing and return your own message instead. I hope you can make some sense of my ramblings 🤣

andersjonsson:
I've created a PR #788 to illustrate my idea. What do you think?
You can clone my branch if you'd like to try it out

moinessim:
Hi Anders,

Thank you for your PR.
I think it's an elegant approach.

I must take some time to test it out soon.

andersjonsson:
Can this PR be closed, now that #788 is merged?

PR #766, opened 2021-12-01
Allow serialization of invalid xml characters

andersjonsson:
Full framework Web Service allows weird control characters etc. in the xml output but XmlTextWriter doesn't by default.
Added Option "CheckXmlCharacters" (default true) to give control over CheckCharacters-property on XmlTextWriter


andersjonsson:
If you merge this PR I'd appreciate a new nuget package

PR #777, opened 2021-12-08
fix issue #575: request parameters are null when using MessageContract

dmolochnikov:


kotovaleksandr:
Hi @dmolochnikov! Thanks for contributing! Please, add test which cover behavior of #575 and be careful with code-style warnings.

dmolochnikov:
> Hi @dmolochnikov! Thanks for contributing! Please, add test which cover behavior of #575 and be careful with code-style warnings.

Aleksandr, thanks for your comments! I'll fix style issues and will try to write a test that checks this specific case

PR #782, opened 2021-12-16
support for extra xmlnamespaceprefixoverrides per messageversion

### LINKED ISSUE #781: XmlNamespaceManager can't handle duplicate prefix
In my legacy service, the response envelope uses the prefix "soap" regardless of what soap version you're running.

So I added both namespaces to my XmlNamespacePrefixOverrides
```
nsm.AddNamespace("soap", "http://www.w3.org/2003/05/soap-envelope");
nsm.AddNamespace("soap", "http://schemas.xmlsoap.org/soap/envelope/");
```

Now I realize that XmlNamespaceManager can't handle duplicate prefixes. So the returnmessages with namespace "http://www.w3.org/2003/05/soap-envelope" gets "prefix3" instead.

andersjonsson:
Fixes #781

Uses the single XmlNamespacePrefixOverrides as default, and adds any MessageVersion specific overrides on top.
XmlNamespaceManager will use the latest namespace for any prefix, so this can be used to override namespaces from the base XmlNamespacePrefixOverrides

If you're OK with this PR I'd appreciate a new nuget version

### DIFF
--- a/src/SoapCore/SoapCoreOptions.cs
+++ b/src/SoapCore/SoapCoreOptions.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.ServiceModel.Channels;
 using System.Xml;
 using SoapCore.Extensibility;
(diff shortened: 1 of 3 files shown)


andersjonsson:
An alternative to the dictionary would be to add the overrides to SoapEncoderOptions instead. That might easier to grasp. What do you think?

andersjonsson:
I’ve done some more thinking and I’ll move the overrides to the soapmessageencoder. Hopefully I can get it done in the weekend

andersjonsson:
Done. Sorry about the conflicts in between this and #785

kotovaleksandr:
@andersjonsson Hi! Please, fix conflicts again:)

andersjonsson:
Done!

PR #788, opened 2021-12-21
Added ISoapMessageProcessor and accompanying unit tests

andersjonsson:
Trying out a concept I've been thinking about.

The idea is to create something like messageinspector, but with the added possibility to "short circuit" the processing and return another message.

Any thoughts?

### DIFF
--- a/src/SoapCore.Tests/SoapMessageProcessor/Startup.cs
+++ b/src/SoapCore.Tests/SoapMessageProcessor/Startup.cs
@@ -1,0 +1,44 @@
+using System.ServiceModel.Channels;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Logging;
+
+namespace SoapCore.Tests.SoapMessageProcessor
+{
+	public class Startup
+	{
+		public void ConfigureServices(IServiceCollection services)
+		{
+			services.AddSoapCore();
+			services.TryAddSingleton<TestService>();
+			services.AddSoapMessageProcessor(async (message, httpcontext, next) =>
+			{
+				await Task.Delay(1);
+
+				if (httpcontext.Request.Path.Value.Contains("ServiceWithProcessor.svc"))
+				{
+					return Message.CreateMessage(MessageVersion.Soap11, "none");
+				}
+				else
+				{
+					return await next(message);
+				}
+			});
+			services.AddMvc();
+		}
+
+		public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILoggerFactory loggerFactory)
+		{
+			app.UseRouting();
+
+			app.UseEndpoints(x =>
+			{
+				x.UseSoapEndpoint<TestService>("/ServiceWithProcessor.svc", new SoapEncoderOptions(), SoapSerializer.DataContractSerializer);
+				x.UseSoapEndpoint<TestService>("/Service.svc", new SoapEncoderOptions(), SoapSerializer.DataContractSerializer);
+			});
+		}
+	}
+}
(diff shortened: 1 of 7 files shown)


src/SoapCore.Tests/SoapMessageProcessor/Startup.cs
```diff
@@ -0,0 +1,44 @@
...
+		{
+			services.AddSoapCore();
+			services.TryAddSingleton<TestService>();
+			services.AddSoapMessageProcessor(async (message, httpcontext, next) =>
+			{
+				await Task.Delay(1);
+
+				if (httpcontext.Request.Path.Value.Contains("ServiceWithProcessor.svc"))
+				{
+					return Message.CreateMessage(MessageVersion.Soap11, "none");
```
moinessim, line 23: I'm getting this error message when I try to `message.WriteBodyContents( <some-writer> )`.

      System.InvalidOperationException: This message cannot support the operation because it has been written.
         at System.ServiceModel.Channels.Message.EnsureWriteMessageState(XmlDictionaryWriter writer)
         at System.ServiceModel.Channels.Message.WriteMessage(XmlDictionaryWriter writer)
         at SoapCore.MessageEncoder.SoapMessageEncoder.WriteMessageAsync(Message message, PipeWriter pipeWriter) in /Users/moisesnessim/code/SoapCore/src/SoapCore/Mess
ageEncoder/SoapMessageEncoder.cs:line 173

andersjonsson: This is when you are trying to read the message and act differently depending on the contents?
Messages are weird and can only be written (read) once. You need to do something like this

```
var bufferedMessage = message.CreateBufferedCopy(int.MaxValue);
var msg = bufferedMessage.CreateMessage();
var reader = msg.GetReaderAtBodyContents();
var content = reader.ReadInnerXml();

return bufferedMessage.CreateMessage();
```

If everything is in order, return `bufferedMessage.CreateMessage();`. Otherwise create and return your own message.

andersjonsson:
Another thing i've been thinking about is to use a return type that contains more than just the Message.

Something like
```
public class SoapResponse
{
    int StatusCode {get;set;}
    Dictionary<string, string> Headers {get;set;}
    Message SoapMessage {get;set;}
}
```
Requiring you to set headers and statuscode directly on the HttpContext could get messy, if multiple messageprocessors add different headers to the response.

What do you think?

src/SoapCore.Tests/SoapMessageProcessor/Startup.cs
moinessim: No, this is when I want to return a new Message, but I want it to have some content, like a serialized DTO, or some other XML.
Apparently, even if I return a new Message and not call "next", something down the pipeline will try to write into the new message. 
And, because I already wrote it, the exception is raised.

andersjonsson: `message.WriteBodyContents( <some-writer> )` writes the content of the message to the supplied writer. Not the other way around.

To create a message with content you have to use one of the overloads of Message.CreateMessage()

I just tried changing my test to:

```
if (httpcontext.Request.Path.Value.Contains("ServiceWithProcessor.svc"))
{
	return Message.CreateMessage(MessageVersion.Soap11, "none", "This is the body");
}
else
{
	return await next(message);
}
```

That worked and gave me 
```
<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/">
  <s:Header>
    <Action s:mustUnderstand="1" xmlns="http://schemas.microsoft.com/ws/2005/05/addressing/none">none</Action>
  </s:Header>
  <s:Body>
    <string xmlns="http://schemas.microsoft.com/2003/10/Serialization/">This is the body</string>
  </s:Body>
</s:Envelope>
```
in return

andersjonsson: If I'm missing something, please add a code sample of what you're trying to do and I'll see if I can figure it out

moinessim: Oh, thanks! 
I actually forgot that overload existed.
Anyway, because my DTO has attributes for the XmlSerializer, I used [CreateMessage(MessageVersion, String, XmlReader)](https://docs.microsoft.com/en-us/dotnet/api/system.servicemodel.channels.message.createmessage?view=dotnet-plat-ext-6.0#System_ServiceModel_Channels_Message_CreateMessage_System_ServiceModel_Channels_MessageVersion_System_String_System_Xml_XmlReader_) , following [this guide](https://weblogs.asp.net/rternier/sending-xml-data-through-wcf-using-system-servicemodel-channels-message-without-using-datacontractserializer) and it worked fine!

andersjonsson:
@kotovaleksandr I would love to hear your thoughts on this.
I'm ready to convert this from draft and get it merged if you like the idea and the implementation.

kotovaleksandr:
Hi @andersjonsson ! Sorry for later response. Looks good, please prepare PR, i will merge them.

andersjonsson:
Done. Let me know if there is anything you would like me to fix before this is merged

PR #801, opened 2022-01-12
Additional namespace declaration attributes in envelope

ZokiPokiMk:
Added new option AdditionalEnvelopeXmlnsAttributes
when:
`
opt.AdditionalEnvelopeXmlnsAttributes = new Dictionary<string, string>()
{
	{"arr","http://schemas.microsoft.com/2003/10/Serialization/Arrays" }
};
`
writes `xmlns:arr="http://schemas.microsoft.com/2003/10/Serialization/Arrays"` in envelope:
`﻿<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:arr="http://schemas.microsoft.com/2003/10/Serialization/Arrays">
`

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,42 @@ public class MyService : IMyServiceService
     }
 }
 ```
+#### Additional namespace declaration attributes in envelope
+Adding additional namespaces to the **SOAP Envelope** can be done by populating `SoapEncoderOptions.AdditionalEnvelopeXmlnsAttributes` parameter.
+```csharp
+....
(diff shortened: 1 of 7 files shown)


kotovaleksandr:
Hi @ZokiPokiMk Thanks for contribution! Please add unit tests and update readme.md

PR #805, opened 2022-01-14
Return XSD instead of WSDL

### LINKED ISSUE #802: External XSD file
Hello, friends!

I am using SoapCore 1.1.0.24 with external WSDL and XSD options. The WSDL part it's working fine, but when I try to access the XSD resource (http://localhost:5180/myservice?xsd&name=myservice.xsd), the WSDL resource is returned instead. This causes that when I import the WSDL on SoapUI the schema part is not been loaded.

hmiguel:
Return XSD instead of WSDL when XSD parameter is present and content type is not defined

Fix #802

### DIFF
--- a/src/SoapCore.Tests/WsdlFromFile/WsdlTests.cs
+++ b/src/SoapCore.Tests/WsdlFromFile/WsdlTests.cs
@@ -4,20 +4,14 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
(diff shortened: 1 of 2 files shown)


kotovaleksandr:
Hi @hmiguel ! Thanks for contribution, please add unit test to PR

hmiguel:
> Hi @hmiguel ! Thanks for contribution, please add unit test to PR

@kotovaleksandr, I changed the existing tests because they were not validating the response content. Is that okay?

PR #819, opened 2022-02-07
Feature: Allow HTTP GET/POST with named endpoints

### LINKED ISSUE #814: Feature: Ability to use the "Simple" requests for GET/POST directly to a method/endpoint
Examples on how they can be called

```plain
GET  /api.asmx/ThisIsAWebserviceMethod?param1=test
POST /api.asmx/ThisIsAPostMethod
```

This is a feature built into .NET Frameworks WebService that can be used for "simple" methods, as I understand it.
The returned data does not contain the SoapEnvelope at all, but directly returns the XML-serialized data.

itssimple:
Fixes #814 

So, basically what this pull request achieves, is allow developers to enable the ability to use "simple" requests to call the API.

Example would be calling `GET /Service.asmx/Ping?s=Hello+world`, which would return:
```xml
<?xml version="1.0" encoding="utf-16"?>
<String xmlns="http://tempuri.org/">Hello world</String>
```

While the normal SOAP API call would return:
```xml
<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <s:Body>
    <PingResponse xmlns="http://tempuri.org/">
      <PingResult>Hello world</PingResult>
    </PingResponse>
  </s:Body>
</s:Envelope>
```

## Changes

- Added an extra route (`/{methodName}`), and then we send those requests through `ProcessHttpOperation`.
- Added option for `HttpPostEnabled`, `HttpsPostEnabled` to enable the users to allow/disallow the use of HTTP POST.
- Added method `TryGetOperation` to remove some duplicate code used to find which operation is being called.
- Added extra methods in the sample service to be able to fix the problem with results not getting their proper type (`ArrayOf<Type>`), methods added `IntArray` and `ComplexReturnModel`
- Added parameter to `ServiceBodyWriter`, to disable envelope generation, and just serialize the result instead.

Co-authored with @andersjonsson


kotovaleksandr:
Great job! Thanks!

PR #831, opened 2022-03-11
fix: If the Request method is HEAD, we will just silently return an OK.

### LINKED ISSUE #830: HEAD request wsdl
`HEAD` requests to the service-wsdl cause the SoapMiddleware to fail

>Request starting HTTP/1.1 HEAD https://...Service1.asmx - -
An error occurred when trying to service a request on SOAP endpoint: /EssenceService5.asmx
Request finished HTTP/1.1 HEAD https://...Service1.asmx - - - 500 0 - 2.7007ms

> System.ArgumentNullException: Value cannot be null. (Parameter 'contentType'
 at SoapCore.MessageEncoder.SoapMessageEncoder.IsContentTypeSupported(String contentType
 at SoapCore.SoapEndpointMiddleware1.ProcessOperation(HttpContext httpContext, IServiceProvider serviceProvider
 at SoapCore.SoapEndpointMiddleware1.Invoke(HttpContext httpContext, IServiceProvider serviceProvider
 at Microsoft.AspNetCore.Routing.EndpointMiddleware.<Invoke>g__AwaitRequestTask|6_0(Endpoint endpoint, Task requestTask, ILogger logger
 at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)

itssimple:
Proposed fix for issue #830 

Since Soap doesn't have any specific implementation of HEAD requests, I suggest that we just silently ignore the request and return `200 OK`, to let the requester know that the address exists at least, and that we do it as early as possible, before any other checks.

fixes #830

### DIFF
--- a/src/SoapCore/SoapEndpointMiddleware.cs
+++ b/src/SoapCore/SoapEndpointMiddleware.cs
@@ -82,9 +82,18 @@ namespace SoapCore
 
 			trailPathTuner?.ConvertPath(httpContext);
 
+			var requestMethod = httpContext.Request.Method;
+
 			if (httpContext.Request.Path.StartsWithSegments(_options.Path, _pathComparisonStrategy, out var remainingPath))
 			{
-				if (httpContext.Request.Method?.ToLower() == "get")
+				if (requestMethod?.ToLower() == "head")
+				{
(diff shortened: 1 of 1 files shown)


kotovaleksandr:
Thanks!

PR #832, opened 2022-03-11
FIX invalid WSDL generation

MeikelLP:
WSDL generator for `DataContractSerializer` couldn't handle duplicate type names in different namespaces

### DIFF
--- a/src/SoapCore/Meta/MetaWCFBodyWriter.cs
+++ b/src/SoapCore/Meta/MetaWCFBodyWriter.cs
@@ -793,7 +793,7 @@ namespace SoapCore.Meta
 		{
 			var toBuildName = GetTypeName(type);
 
-			if (_builtComplexTypes.Contains(toBuildName))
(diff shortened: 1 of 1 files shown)


andersjonsson:
Hi @MeikelLP 
Thanks for submitting a PR!

I'm probably missing something, but won't the wsdl be invalid with two types of the same name?

MeikelLP:
@andersjonsson No that is fine because both are put in their respective namespace. The only issue was that the "we already processed that type" cache-list wasn't respecting the namespace.

andersjonsson:
Great, thanks!

MeikelLP:
Additional note for everybody who stumbles across this issue again.

Even with this fix I still had this issue to some degree. I don't know if this PR is helping at all. But I think it doesn't hurt either. 

Anyway be sure to to set explicit Soap `Namespace`s in your types which use `DataContractAttribute`. This resolved it for us at least.

PR #841, opened 2022-03-19
Set Content-Length header in response

acveigas:
Came across few SOAP clients which require **Content-Length** header to be set in order for the SOAP request to go through. 
Without this change, **Content-Length** header is not set and **Transfer-Encoding** is **chunked**


acveigas:
Working on squashing the commits

PR #843, opened 2022-03-21
Set Content-Length header in response

acveigas:
Came across few SOAP clients which require **Content-Length** header to be set in order for the SOAP request to go through.
Without this change, **Content-Length** header is not set and **Transfer-Encoding** is **chunked**

### DIFF
--- a/src/SoapCore/CustomStringWriter.cs
+++ b/src/SoapCore/CustomStringWriter.cs
@@ -1,0 +1,23 @@
+using System.IO;
+using System.Text;
+
+namespace SoapCore
+{    
+    public class CustomStringWriter : StringWriter
+    {
+        private readonly Encoding _encoding;
+
+		public CustomStringWriter(Encoding encoding)
+		{
+			_encoding = encoding;
+		}
+
+        public override Encoding Encoding
+        {
+            get
+            {
+                return _encoding;
+            }
+        }        
+    }
+}
--- a/src/SoapCore/MessageEncoder/SoapMessageEncoder.cs
+++ b/src/SoapCore/MessageEncoder/SoapMessageEncoder.cs
@@ -147,7 +147,7 @@ namespace SoapCore.MessageEncoder
 			return Task.FromResult(message);
 		}
 
-		public virtual async Task WriteMessageAsync(Message message, HttpContext httpContext)
+		public virtual async Task WriteMessageAsync(Message message, HttpContext httpContext, PipeWriter pipeWriter)
 		{
 			if (message == null)
 			{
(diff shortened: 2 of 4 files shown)


acveigas:
@andersjonsson Can you approve the actions?

andersjonsson:
Thanks for the PR!
I’ll take a closer look later today

acveigas:
Thanks. I Will also look at the CodeQL failure later on this week.

> Thanks for the PR!
> 
> I’ll take a closer look later today

src/SoapCore/MessageEncoder/SoapMessageEncoder.cs
```diff
@@ -146,35 +147,53 @@ public Task<Message> ReadMessageAsync(Stream stream, int maxSizeOfHeaders, strin
...
 
-			using var xmlWriter = XmlDictionaryWriter.CreateDictionaryWriter(xmlTextWriter);
-			message.WriteMessage(xmlWriter);
-			xmlWriter.WriteEndDocument();
-			xmlWriter.Flush();
+				var data = stringWriter.ToString();
 
-			await pipeWriter.FlushAsync();
+				//Set Content-length in Response
+				httpContext.Response.ContentLength = data.Length;
```
andersjonsson, line 191: Do these legacy clients expect ContentLength to be the number of characters in the message, or the number of bytes?

src/SoapCore/CustomStringWriter.cs
```diff
@@ -0,0 +1,23 @@
+using System.IO;
+using System.Text;
+
+namespace SoapCore
+{    
+    public class CustomStringWriter : StringWriter
```
andersjonsson, line 6: Would it be possible to create a TextWriter that writes directly to the pipewriter, while keeping track of how much data has been written?
The perf is probably fine as it is, but if we can avoid unnecessary allocations and copying of data it's a good thing 😉

andersjonsson (commented):
Left a few notes.

src/SoapCore/MessageEncoder/SoapMessageEncoder.cs
acveigas: The legacy client worked with number of characters. Tested out with a unit test case and both the lengths were same. Leaving this change as is for now.

src/SoapCore/CustomStringWriter.cs
acveigas: @andersjonsson Tried this approach but something keeps breaking. Either its my unit testing or the automated tests. I guess we'll have to proceed this way.

acveigas:
I can send a PR with number of bytes. Shouldn't take time 😊

PR #848, opened 2022-03-25
fix: Don't output the BOM, it can crash clients.

### LINKED ISSUE #847: Unwanted/Risky UTF8 Byte Order marks at the start of the http responses.
Further to minor glitches in some client code, I used Telerik Fiddler (HexView mode), to confirm that **SoapCore is generating UTF8 Byte Order Marks at the very beginning of the http response bodies**.

This is useless and possibly harmful.

The case that I noticed can be easily fixed in [SoapEncoderOptions.cs](https://github.com/DigDes/SoapCore/blob/3f55bb9fcae723a7e73c9141779bb3e31a6bef50/src/SoapCore/SoapEncoderOptions.cs#L10) replacing 
`Encoding.UTF8` by `new UTF8Encoding(encoderShouldEmitUTF8Identifier: false)` .

This being said, there are plenty of other occurences of "Encoding.UTF8" in the code.. some are not a problem like `Encoding.UTF8.GetBytes(string)`, others might be.

itssimple:
fixes #847

### DIFF
--- a/src/SoapCore/SoapEndpointMiddleware.cs
+++ b/src/SoapCore/SoapEndpointMiddleware.cs
@@ -391,7 +391,7 @@ namespace SoapCore
 			context.Response.ContentType = "text/xml";
 
 			using var ms = new MemoryStream();
-			XmlWriter writer = XmlWriter.Create(ms, new XmlWriterSettings() { Encoding = new UTF8Encoding(true) });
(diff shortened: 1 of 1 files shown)


andersjonsson:
Competing PR's 😆 
Will close this as mine covers all instances of UTF8 encodings used

PR #856, opened 2022-03-29
Add interface

psvetlov:
fixes #856

### DIFF
--- a/src/SoapCore/CustomMessage.cs
+++ b/src/SoapCore/CustomMessage.cs
@@ -52,6 +52,11 @@ namespace SoapCore
 			}
 		}
 
+		protected override void OnWriteStartHeaders(XmlDictionaryWriter writer)
+		{
+			writer.WriteStartElement(Version.Envelope.NamespacePrefix(NamespaceManager), "Header", Version.Envelope.Namespace());
+		}
+
 		protected override void OnWriteStartBody(XmlDictionaryWriter writer)
 		{
 			writer.WriteStartElement(Version.Envelope.NamespacePrefix(NamespaceManager), "Body", Version.Envelope.Namespace());


src/SoapCore/TrailingServicePathTuner.cs
```diff
@@ -7,11 +7,16 @@
 
 namespace SoapCore
 {
+    public interface ITrailingServicePathTuner
```
andersjonsson, line 10: What do you think about calling it IServicePathTuner? The trailing-bit could be considered a behavior of the default implementation

src/SoapCore/SoapEndpointMiddleware.cs
```diff
@@ -79,8 +79,8 @@ public SoapEndpointMiddleware(ILogger<SoapEndpointMiddleware<T_MESSAGE>> logger,
 
 		public async Task Invoke(HttpContext httpContext, IServiceProvider serviceProvider)
 		{
-			var trailPathTuner = serviceProvider.GetService<TrailingServicePathTuner>();
-
+			var trailPathTuner = serviceProvider.GetService<ITrailingServicePathTuner>();
```
andersjonsson, line 82: There is a failing unit test connected to this change. In it's current form this would be a breaking change.
```suggestion
			var trailPathTuner = serviceProvider.GetService<ITrailingServicePathTuner>() ??serviceProvider.GetService<TrailingServicePathTuner>();

```
With this change I _guess_ that we won't break existing code

andersjonsson (commented):
Thank you for the contribution! 
I left a couple of comments

src/SoapCore/TrailingServicePathTuner.cs
psvetlov: You are right! PathString is better!

src/SoapCore/SoapEndpointMiddleware.cs
psvetlov: Sorry I haven't run any tests but I would say if it requires non-null when it should be GetRequiredService... meantime I see checking for null later... 'TrailPathTuner?.ConvertPath(httpContext)'
It seems the unit tests are not correct. (And please note that for Unity DI it will always be non-null with your suggestion and the only way is to create a dummy tuner)

andersjonsson: Sorry, I didn't explain this well enough.

Current usage depends on people registering TrailingServicePathTuner. A call to `serviceProvider.GetService<ITrailingServicePathTuner>()` won't hit that registration.
The code won't crash (it handles null) but the TrailingServicePathTuner won't be invoked, meaning that the behavior of existing services will change (if it depends on the tuner, that is).

andersjonsson: With my change the result will be null if you haven't registered a tuner at all

andersjonsson: An alternative to this fix is to rename TrailingServicePathTuner. That way people using it will get a compile time error and will (hopefully) investigate enough to find the interface.
What do you think?

psvetlov: 1) I agree, this can be overridden in a derived class that can be used for registration to provide custom path behavior. That fixes all problems.
2) Some Di frameworks (my case) can instantiate even an unregistered class. So the result could be not obvious. For this case, the interface is more suitable.

andersjonsson:
Thanks again for this contribution. I'll give this another review as soon as you've had the time to fix the unit test and address the design issues we discussed

Issue #863, opened 2022-04-08
Is it possible to invoke the service method by reading it from the soap header?


Blackleones:
The soap message that I am trying to consume does not have the action name inside the envelope's body but it is described in the `<action>` inside the header.

Is there a possible way on how to let soapcore determines the method to invoke reading it from the action tag inside the soap header?

```xml
<soap:Envelope xmlns:soap="http://www.w3.org/2003/05/soap-envelope" xmlns:wsa="http://www.w3.org/2005/08/addressing">
    <soap:Header>
        <wsa:Action>action_to_trigger</wsa:Action>
        <wsa:MessageID>message_id</wsa:MessageID>
    </soap:Header>
    <soap:Body>
        ...
    </soap:Body>
</soap:Envelope>    
```

Blackleones:
Actually, I have found a workaround using another middleware in front of SoapCore but I wonder if there is a better solution 

```csharp
public class CustomPreSoapMiddleware
{
    private readonly RequestDelegate _next;
    private const string InterestedPath = "/Service.asmx";
    private const string InterestedTag = "wsa:Action";
    public CustomPreSoapMiddleware(RequestDelegate next)
    {
        _next = next;
    }

    public async Task InvokeAsync(HttpContext context)
    {
        var req = context.Request;
        
        if (req.Method.ToLower() == "post" && req.Path.Equals(InterestedPath))
        {
            req.EnableBuffering();

            using var reader = new StreamReader(req.Body, encoding: Encoding.UTF8, detectEncodingFromByteOrderMarks: false, leaveOpen: true);
            
            var body = await reader.ReadToEndAsync();
            
            XmlDocument xd = new XmlDocument();
            xd.LoadXml(body);

            TrySetSoapAction(req, xd);
            
            req.Body.Position = 0;
        }
        
        await _next(context);
    }

    private bool TrySetSoapAction(HttpRequest request, XmlDocument xd)
    {
        var action = xd.GetElementsByTagName(InterestedTag).Cast<XmlNode>().FirstOrDefault();

        if (action is not null && !string.IsNullOrEmpty(action.InnerText))
        {
            request.Headers["SOAPAction"] = action.InnerText;
            return true;
        }

        return false;
    }
}
```

andersjonsson:
For now the workaround seems to be the only way.
Shouldn't be too hard to add though, if you feel up for adding this functionality in a PR
Take a look at
[...]

PR #866, opened 2022-04-16
Read SOAPAction from soap envelope's headers

Blackleones:
Actually, it is not possible to retrieve the SOAPAction from the soap request's header. This code tries to check if the soap envelope contains any reference about what is the SOAPAction inside the header Action tag. If so, then it read the SOAPAction from the found tag.

issue: #863

About tests:

I've noticed that the following tests:

    Soap12Wsa10Header
    Soap12PingWithActionInEnvelopeHeader
    Soap12Wsa10FaultHeader

Supposed to let SoapCore retrieves the SOAPAction from the header but it retrieved that from the body. Now, this PR lets these tests work properly.

### DIFF
--- a/src/SoapCore/HeadersHelper.cs
+++ b/src/SoapCore/HeadersHelper.cs
@@ -98,6 +98,24 @@ namespace SoapCore
 					}
 				}
 
+				if (string.IsNullOrEmpty(soapAction))
+				{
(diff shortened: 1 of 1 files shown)


andersjonsson:

Looks good 👍

PR #868, opened 2022-04-27
Add support for arrays with choice

gaetan-lebris-shift:
Linked to issue #869 

My fix is not complete, it can only deserialize a full array of A, B or C, not a mix of the three but at the time, I am okay with it.
(plus it would require adding a loop or a rethink of the ArrayDeserializing method ..)

I am new to contributing and to Soap, please tell me if anything is wrong. :)


kotovaleksandr:
Hi @gaetan-lebris-shift ! Yes, please create issue with your case, after that i will merge PR.

gaetan-lebris-shift:
Hello @kotovaleksandr , thank you for your swift reply. I just created an issue.

PR #874, opened 2022-05-11
#873 Operation cannot be found issue

### LINKED ISSUE #873: Operation cannot be found issue
Hi to all,
We have a problem if the header soapaction is not specified. When soapaction is not specified then middlware cannot find which service method to call. 

Method name: 
image

I found a solution so I ask if you could add that to the code.
Problem is in TryGetOperation method. The picture shows the as is state and that is the state with the change. After the change, there is no problem anymore. My question is if you could add that in the next version.

image

kristijanmihaljinac:
fixes #873


andersjonsson:
Nice, thanks!

Issue #872, opened 2022-05-06
UsernameToken issue


evgbob:
Hi to all.
Why I can't use such header:
```
<wsse:UsernameToken>
         <wsse:Username>Zoe</wsse:Username>
         <wsse:Password>IloveDogs</wsse:Password>
         <wsse:Nonce EncodingType="...#Base64Binary">MTY1MDk4NDg1Nzk3MA==</wsse:Nonce>
</wsse:UsernameToken>
```

In source code first check is:
```
if (wsUsernameToken.Nonce != null ^ wsUsernameToken.Created != null) //// true here because "Created" is missing
{
      throw new Exception();
}
```

And only after that:
```
var isClearText = wsUsernameToken.Password?.Type == null || wsUsernameToken.Password.Type == _passwordTextType;
if (isClearText)
{
      return wsUsernameToken.Password?.Value == _password;
}
```
In my case "isClearText" will be true, so why need first check. I think need to swap them, because Nonce and Created is optional attributes ([page 10](http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-username-token-profile-1.0.pdf))
Nonce and Created must be used only if Password type is PasswordDigest.
Thanks.

andersjonsson:
If Nonce and Created are optional, we shouldn't need to check for them at all, right?

Feel like submitting a PR?

evgbob:
Already done.

PR #876, opened 2022-05-11
Nonce and Created header attributes uses for check only if password h…

evgbob:
Nonce and Created header attributes uses for check only if password has not type PasswordText.
issue #872


andersjonsson:
Thanks!

PR #877, opened 2022-05-18
Added SOAP 1.2 Support using the DataContractSerializer

### LINKED ISSUE #825: SOAP 1.2 Support with DataContractSerializer not working
Hi

I am try to setup support for SOAP 1.2 using SoapCore with the DataContractSerializer. I am not sure if I am doing something wrong, however based on previous SoapCore versions when setting up the MessageVersion in the SoapEncoderOption the appropriate version of SOAP would be supported. When I am selecting any of the SOAP 1.2 flavours (Soap12WSAddressing10 or Soap12WSAddressingAugust2004) I still get SOAP 1.1 in the WSDL. This is not the issue if I use the XmlSerializer. 

Am I missing some configuration or is this an issue? I have attached a demo project.

[SOAPTester.zip](https://github.com/DigDes/SoapCore/files/8117163/SOAPTester.zip)

GreekOctopus:
- Updated the MetaWCFBodyWritter constructor (DataContractSerializer) to accept the SoapBindings as a parameter so we can establish which version of SOAP to use. Created supporting methods as per the MetaBodyWriter class.
- Added the missing WS-Addressing 1.0 WSDL Binding Namespace in the Namespaces.
- Moved parts of the MetaMessage outside the check for BasicAuthentication, as I believe they need to be always present.
- Updated SoapEndpointMiddleware to use the dynamic binding names instead of the hard coded HttpBinding.
- Updated all relevant tests to test using both serializers. Also moved the hard coded vales for the Binding/Port to constants.

Fixes #825

### DIFF
--- a/src/SoapCore/SoapEndpointMiddleware.cs
+++ b/src/SoapCore/SoapEndpointMiddleware.cs
@@ -233,11 +233,11 @@ namespace SoapCore
 		{
 			var baseUrl = httpContext.Request.Scheme + "://" + httpContext.Request.Host + httpContext.Request.PathBase + httpContext.Request.Path;
 			var xmlNamespaceManager = GetXmlNamespaceManager(null);
-			var bindingName = "BasicHttpBinding_" + _service.GeneralContract.Name;
+			var bindingName = _options.EncoderOptions[0].BindingName + "_" + _service.GeneralContract.Name;
 
 			var bodyWriter = _options.SoapSerializer == SoapSerializer.XmlSerializer
 				? new MetaBodyWriter(_service, baseUrl, xmlNamespaceManager, bindingName, _messageEncoders.Select(me => new SoapBindingInfo(me.MessageVersion, me.BindingName, me.PortName)).ToArray())
-				: (BodyWriter)new MetaWCFBodyWriter(_service, baseUrl, bindingName, _options.UseBasicAuthentication);
+				: (BodyWriter)new MetaWCFBodyWriter(_service, baseUrl, bindingName, _options.UseBasicAuthentication, _messageEncoders.Select(me => new SoapBindingInfo(me.MessageVersion, me.BindingName, me.PortName)).ToArray());
 
 			//assumption that you want soap12 if your service supports that
 			var messageEncoder = _messageEncoders.FirstOrDefault(me => me.MessageVersion == MessageVersion.Soap12WSAddressing10 || me.MessageVersion == MessageVersion.Soap12WSAddressingAugust2004) ?? _messageEncoders[0];
--- a/src/SoapCore/SoapEndpointMiddleware.cs
+++ b/src/SoapCore/SoapEndpointMiddleware.cs
@@ -233,8 +233,7 @@ namespace SoapCore
 		{
 			var baseUrl = httpContext.Request.Scheme + "://" + httpContext.Request.Host + httpContext.Request.PathBase + httpContext.Request.Path;
 			var xmlNamespaceManager = GetXmlNamespaceManager(null);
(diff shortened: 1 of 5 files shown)


src/SoapCore/SoapEndpointMiddleware.cs
```diff
@@ -233,11 +233,11 @@ private async Task ProcessMeta(HttpContext httpContext, bool showDocumentation)
 		{
 			var baseUrl = httpContext.Request.Scheme + "://" + httpContext.Request.Host + httpContext.Request.PathBase + httpContext.Request.Path;
 			var xmlNamespaceManager = GetXmlNamespaceManager(null);
-			var bindingName = "BasicHttpBinding_" + _service.GeneralContract.Name;
+			var bindingName = _options.EncoderOptions[0].BindingName + "_" + _service.GeneralContract.Name;
```
andersjonsson, line 236: Since this variable will be set here, even if EncoderOptions.BindingName is null, I don't think the fallback, to use "BasicHttpBinding" as default name, works. 
Could you verify that the meta document is the same as before this PR, by default?

andersjonsson:
I left a comment that I hope you can look into

GreekOctopus:
> Thank you for your contribution! I left a comment that I hope you can look into

I'm sorry but I cannot find the comment. I looked through all the files and I cannot see any changes/comments. As I am new to Github would you be kind enough to point me to the right direction?

andersjonsson:
> I'm sorry but I cannot find the comment. I looked through all the files and I cannot see any changes/comments. As I am new to Github would you be kind enough to point me to the right direction?

Sorry, I forgot to actually finish my code review 🤦‍♂️ 

I just want you to make sure that the default values stay the same. Wouldn't want to introduce breaking changes.
As soon as that bit is cleared I think we can merge this PR in.

src/SoapCore/SoapEndpointMiddleware.cs
GreekOctopus: Would you be happy to read from the EncoderOptions but if they are empty we use the deault value?
Something like this:

var bindingName = !string.IsNullOrWhiteSpace(_options.EncoderOptions[0].BindingName) ? _options.EncoderOptions[0].BindingName : "BasicHttpBinding_" + _service.GeneralContract.Name;

GreekOctopus: Regarding the meta body if I understand this correctly, you are referring to the non web service specific fields. I can send you a copy to reconfirm. They are identical apart from now we have the soap12 tags instead of soap as expected and I have included a Policy with WsHttpBinding. When using a BasicBinding and Soap 1.1 the meta body is identical. If I did not understand your question, if you do not mind explaining a little bit more I am happy to check exactly what you need and get back to you.

andersjonsson: Yep, your suggested change for binding name seems correct. 
Just make sure that if one doesn't set any options the binding name etc. stays the same. Then I'm happy.

GreekOctopus: I just updated the PR. I updated how we handle a default value if none is present. I updated the relavent tests. Checked the metadata for SOAP 1.1 before and after the PR and it seems correct. I removed couple of hard coded sections (PolicyReference, UsingAddressing) as I wrongly assumed that they were always needed.

andersjonsson:
Thanks!

GreekOctopus:
> Thanks!

Thank you for approving and merging. Is there any chance to bring the next release out sooner rather than later? I obviously do not know what other work is being done in the background, but I do need an official release to be consumed in my current work project. Thank you

PR #879, opened 2022-05-26
Default MaxSoapHeaderSize of 0x10000 moved to SoapEncoderOptions so it can be overridden

it4-igorkn:
Default stays the same.

We have some users on test environment with SAML tokens > 64kB so need to up the limit here.


it4-igorkn:
Thanks for accepting PR.
When can we expect new release?

kotovaleksandr:
@it4-igorkn Done! Thanks!
https://github.com/DigDes/SoapCore/releases/tag/v1.1.0.30
nuget.org currently index package, please wait:)

kotovaleksandr:
https://www.nuget.org/packages/SoapCore/1.1.0.30

PR #904, opened 2022-09-26
Wsdl schema mapping improvements

Aleksanderis:
Added some improvements for external WSDL schemas support:

- `WsdlFileOptionsCaseInsensitive` options - to use with case-insensitive mapping. Without it, even if you use `SoapCoreOptions.CaseInsensitivePath = true` - external WSDL schemas would fail to map by `HttpContext.Request.Path`.
- `UrlOverride` property to `WebServiceWSDLMapping`. It's needed to be able to host services under different folders. For example if you have one service hosted like - https://localhost:5001/Management/Service.asmx, and another one - as https://localhost:5001/Orders/Customers.asmx. I didn't manage to achieve this goal with existing `UrlOverride`, `VirtualPath`, `AppPath` properties in `WsdlFileOptions`.

**Short afterword:** I hope the goals I was trying to achieve are clear. As about implementation - I'm ready to do additional changes in case someone has any better ideas or if there are some general issues or inconsistencies with the nature of SoapCore.
For now I tried to be cautious and add functionality on top. If it would be mine own project - then, for example I perhaps `WsdlFileOptionsCaseInsensitive` could just replace `WsdlFileOptions`, because it's just would be more flexible. However I didn't want to introduce any potential breaking changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,22 +78,26 @@ To use it, add a setting like this to appsettings
 ```csharp
  "FileWSDL": {
     "UrlOverride": "",
+    "VirtualPath": "",
     "WebServiceWSDLMapping": {
-      "Service.asmx": {
+      "Service.asmx": { ,
+        "UrlOverride": "Management/Service.asmx",
         "WsdlFile": "snapshotpull.wsdl",
         "SchemaFolder": "Schemas",
         "WsdlFolder": "Schemas"
       }
-    },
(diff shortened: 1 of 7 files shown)


Aleksanderis:
Force-pushed same commits, but with removed line-breaking changes

andersjonsson (approved):
Looks good!

PR #907, opened 2022-10-18
Add support for DefaultValueAttributes

### LINKED ISSUE #906: Set minOccurs="0" for optional boolean fields with XmlSerializer
Is there any way to specify minOccurs="0" for a boolean field when using the XmlSerializer? I need to add an optional field to an existing API but the field is always generated with minOccurs="1". I can't use the DataContractSerializer as it doesn't work with existing clients.

.NET Framework provided two ways ([MinOccurs Attribute Binding Support](https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-3.5/zds0b35c(v=vs.90))) to do this but neither method seems to work SoapCore.

- Value type with a default value specified via a [System.Component.DefaultValueAttribute](https://msdn.microsoft.com/en-us/library/2cws2s9d(v=vs.90)).
- Value type accompanied by a public bool field that uses the Specified naming convention described previously under Translating XSD to source.

coxp:
Add support for DefaultValueAttribute on fields when using the XmlSerializer. Fixes #906

### DIFF
--- a/src/SoapCore/Meta/MetaBodyWriter.cs
+++ b/src/SoapCore/Meta/MetaBodyWriter.cs
@@ -877,16 +877,16 @@ namespace SoapCore.Meta
 			else
 			{
 				string defaultValue = null;
-				var defaultAttribute = member.GetCustomAttribute<DefaultValueAttribute>();
-				if (defaultAttribute != null)
+				var defaultAttributeValue = member.GetCustomAttribute<DefaultValueAttribute>()?.Value;
+				if (defaultAttributeValue != null)
 				{
-					if (defaultAttribute.Value is bool value)
+					if (defaultAttributeValue is bool value)
 					{
 						defaultValue = value ? "true" : "false";
 					}
 					else
 					{
-						defaultValue = defaultAttribute.Value.ToString();
+						defaultValue = defaultAttributeValue.ToString();
 					}
 				}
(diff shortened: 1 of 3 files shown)


src/SoapCore/Meta/MetaBodyWriter.cs
```diff
@@ -875,7 +876,20 @@ private void AddSchemaTypePropertyOrField(XmlDictionaryWriter writer, MemberInfo
...
+				var defaultAttribute = member.GetCustomAttribute<DefaultValueAttribute>();
+				if (defaultAttribute != null)
+				{
+					if (defaultAttribute.Value is bool value)
+					{
+						defaultValue = value ? "true" : "false";
+					}
+					else
+					{
+						defaultValue = defaultAttribute.Value.ToString();
```
andersjonsson, line 889:
```suggestion
						defaultValue = defaultAttribute.Value?.ToString();

```
Value can be null

andersjonsson (changes requested):

Left a comment about a possible NullReferenceException, but it looks good otherwise

src/SoapCore/Meta/MetaBodyWriter.cs
coxp: Good spot :).

I've made a change to match the behaviour in .NET framework ASMX services. It looks like [DefaultValue(null)] is equivalent to not supplying the attribute.

andersjonsson:
Thanks again!

PR #910, opened 2022-10-20
Remove CancellationToken for async methods from wsdl

ArtemevAL:
Avoid adding CancellationToken in method parameters for asynchronous methods and adding type to type declarations.


andersjonsson:
Thank you!

PR #911, opened 2022-10-20
Remove import contract namespace as this results in a stack overflow

ArtemevAL:
Exclude contract namespace import, as this results in a stack overflow error in some parsers.

Example, before:
...
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:ser="http://schemas.microsoft.com/2003/10/Serialization/" elementFormDefault="qualified" targetNamespace="http://iteco.ru/esz/EszServer/MpguSearchService">
			<xs:import namespace="http://iteco.ru/esz/EszServer/MpguSearchService" />
			<xs:import namespace="http://schemas.microsoft.com/2003/10/Serialization/Arrays"/>
...

after :
...
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:ser="http://schemas.microsoft.com/2003/10/Serialization/" elementFormDefault="qualified" targetNamespace="http://iteco.ru/esz/EszServer/MpguSearchService">
			<xs:import namespace="http://schemas.microsoft.com/2003/10/Serialization/Arrays"/>
...

### DIFF
--- a/src/SoapCore/Meta/MetaWCFBodyWriter.cs
+++ b/src/SoapCore/Meta/MetaWCFBodyWriter.cs
@@ -308,7 +308,7 @@ namespace SoapCore.Meta
 
 			var groupedByNamespace = _complexTypeToBuild.GroupBy(x => x.Value).ToDictionary(x => x.Key, x => x.Select(k => k.Key));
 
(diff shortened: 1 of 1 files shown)


andersjonsson:
I don't have any experience with `MetaWCFBodyWriter` but as far as I can tell this should be fine

Issue #918, opened 2022-11-04
Non-utf-8 encodings are not fully working


Aleksanderis:
We have a legacy clients using our webservices, these clients were not written with utf-8 in mind. Therefore they don't work if `SoapEncoderOptions` having `WriteEncoding = Encoding.UTF8` (or using default encoding, which is also utf-8).

Previously there was already some fix for that - [Support for ISO-8859-1 #429](https://github.com/DigDes/SoapCore/pull/429), but it's seems not enough. If client sends a requests which has some special characters (like Swedish ones - `ö` or `å`) - requests are failing with errors like: 
>_Invalid character in the given encoding. Line 17, position 12._

What I found so far is this quite old discussion - https://social.msdn.microsoft.com/Forums/sqlserver/en-US/267fbe9a-dd78-4c68-9840-d0467fdbd456/xmlexception-invalid-character-in-the-given-encoding

Looks like, if you are not working with utf-8 - encoding should be provided more specifically when `XmlReader.Create()` is called. I'm working on a possible fix, didn't get a proper way to fix it yet, but I tried some dirty workaround for now, to make sure it's the case.

On that line:
https://github.com/DigDes/SoapCore/blob/master/src/SoapCore/MessageEncoder/SoapMessageEncoder.cs#L146

If I just create another StreamReader with same encoding which was specified for `SoapEncoderOptions` everything starts working correctly:

```csharp
reader = XmlReader.Create(new StreamReader(stream, _writeEncoding), xmlReaderSettings);
```

It's just a quick fix for now and it's probably not a good way to create one more StreamReader on top, but initial StreamReader is coming from either from `PipeReader` (from `System.IO.Pipelines`) or from `HttpContext.Request.Body` (which is `Stream`). I will investigate further, but if anyone can come with some faster suggestion how to fix it quickly and correctly - you are welcome. :)

Aleksanderis:
#919 fix wasn't fully enough. Not sure how it was controlled in old ASMX services, but legacy client were able to use different encodings/charsets/contenttypes for request and response.
I.e. client could send a request with Windows-1252 encoding, and then expect to get a response with UTF-8. In that case response content type should be same as response encoding, not as request encoding.

PR #923, opened 2022-11-08
Refactored encoding used for reading to respect ContentType header

Aleksanderis:
It's one more stabilization fix for legacy non-utf-8 clients, started in #918.

It's a refactoring of changes done in #921. Actually it was a bad idea to set a fixed `readEncoding`, it should respect what client is requesting. One client can work with windows-1252/ISO-8859-1, another client might work with utf-8. The way to determine it - is "ContentType" header sent with the request from the client.
So I refactored encoding used for reading to respect "ContentType" header in the request, which should have charset/encoding in it (example - `Content-Type: text/xml; charset=windows-1252`).
If it's not present or some faulty one - it should fallback to previous behavior or just take default encoding (which is utf-8).

### DIFF
--- a/src/SoapCore.Tests/Startup.cs
+++ b/src/SoapCore.Tests/Startup.cs
@@ -136,7 +136,6 @@ namespace SoapCore.Tests
 				{
 					MessageVersion = MessageVersion.Soap11,
 					WriteEncoding = Encoding.UTF8,
-					ReadEncoding = Encoding.GetEncoding("ISO-8859-1"),
 					OverwriteResponseContentType = false
 				};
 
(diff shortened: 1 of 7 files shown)


Aleksanderis:
@kotovaleksandr or @andersjonsson any chance you could review it soon please? :)

kotovaleksandr:
@Aleksanderis need new nuget package?)

Aleksanderis:
> Aleksandras T. need new nuget package?)
@kotovaleksandr yes, it would be lovely. ;)

kotovaleksandr:
https://www.nuget.org/packages/SoapCore/1.1.0.36

Issue #937, opened 2023-03-30
Method parameter's namespace of Array type in WSDL is incorrect


dengber:
The following WSDL generated by SoapCore is incorrect,  line 9 should be ```<xs:element xmlns:q1="http://Test.SoapCore.Services.Dto" minOccurs="1" name="inDtos" nillable="true" type="q1:ArrayOfInDto" />``` .

Test code for this WSDL is [here](https://github.com/dengber/Test.SoapCore)

```
1. <wsdl:definitions xmlns:soap="http://schemas.xmlsoap.org/wsdl/soap/" xmlns:tns="http://Test.SoapCore.Services/Test" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:http="http://schemas.microsoft.com/ws/06/2004/policy/http" xmlns:msc="http://schemas.microsoft.com/ws/2005/12/wsdl/contract" xmlns:wsp="http://schemas.xmlsoap.org/ws/2004/09/policy" xmlns:wsu="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd"
2.     xmlns:wsam="http://www.w3.org/2007/05/addressing/metadata" xmlns:wsdl="http://schemas.xmlsoap.org/wsdl/" targetNamespace="http://Test.SoapCore.Services/Test" name="ITestWebServiceProvider">
3.     <wsdl:types>
4.         <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:ser="http://schemas.microsoft.com/2003/10/Serialization/" elementFormDefault="qualified" targetNamespace="http://Test.SoapCore.Services/Test">
5.             <xs:import namespace="http://Test.SoapCore.Services.Dto" />
6.             <xs:element name="Execute">
7.                 <xs:complexType>
8.                     <xs:sequence>
9.                         <xs:element minOccurs="1" name="inDtos" nillable="true" type="tns:ArrayOfInDto" />
10.                     </xs:sequence>
11.                 </xs:complexType>
12.             </xs:element>
13.             <xs:element name="ExecuteResponse">
14.                 <xs:complexType>
15.                     <xs:sequence>
16.                         <xs:element xmlns:q1="http://Test.SoapCore.Services.Dto" minOccurs="1" name="ExecuteResult" nillable="true" type="q1:ArrayOfOutDto" />
17.                     </xs:sequence>
18.                 </xs:complexType>
19.             </xs:element>
[...]

PR #939, opened 2023-03-31
Fix method parameter's namespace of Array type in WSDL 

dengber:
Fix method parameter's namespace of Array type in WSDL is incorrect problem #937.


andersjonsson:
Thanks!

PR #949, opened 2023-06-27
More namespace flexibility

programmersommer:
For back compatibility, it might be needed to have an empty xmlns="" for the first tag.
This change will allow it to do in case if namespace is specified and set to empty string (Namespace="")
In case it is not specified or set to null would be displayed service namespace (like it is done currently)


andersjonsson:
Seems to be in line with expected behavior. Thank you!

Issue #908, opened 2022-10-19
XmlNode argument always null


shustrenko:
Hi, I am trying to create the following service which expects to receive an envelope body as an argument. The problem that infoReq argument always comes null in the service method. What can the problem here?

```
[ServiceContract]
public interface ITestService
{
    [OperationContract]
    public XmlNode GetInfo(XmlNode infoReq);
}

public class GenericService : IGenericService
{
    public XmlNode GetInfo(XmlNode infoReq)
    {
        return infoReq;
    }
}
```

using XmlSerializer
```
endpoints.UseSoapEndpoint<ITestService>("/TestService.asmx", new SoapEncoderOptions(), SoapSerializer.XmlSerializer);
```

sample of xml which i pass from SoapUI

```
<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:tem="http://tempuri.org/">
   <soapenv:Header/>
   <soapenv:Body>
      <tem:GetInfo>
         <!--Optional:-->
         <tem:infoReq>
            <test>string</test>
            <tem:Object/>
         </tem:infoReq>
      </tem:GetInfo>
   </soapenv:Body>
</soapenv:Envelope>
```

andersjonsson:
My guess is that the deserialization in SoapCore simply can't handle deserialization to XmlNode.

If I were you I'd fork the repo and set up a test case for that. That way you can debug into the deserialization and see what happens.
You are very welcome to ask specific questions if you run into issues, and a PR with a fix would be appreciated!

This is the way I got started contributing to SoapCore. I had a need for specific serialization and wsdl-fixes, so I added them myself and sent PR's

shustrenko:
I was able to reproduce the issue with a next sample code. It uses almost same approach what method SoapCore.SoapEndpointMiddleware.GetRequestArguments does. 
With empty namespace parameterNs ="" deserializer returns not null value. But if i set not empty namespace that Deserializer returns awlays null. Any idea?


```
        private static void TestDeserialization()
        {
            var parameterNs = "http://mynamespace/";
            string inputXml;
            if (string.IsNullOrEmpty(parameterNs))
            {
                inputXml = "<GetInfo>";
            }
            else
            {
                inputXml = $@"<GetInfo xmlns=""{parameterNs}"">";
            }

            inputXml += @"
<infoReq>
<MyClass>
<child1>foo</child1>
<child2>bar</child2>
</MyClass>
</infoReq>
[...]

PR #955, opened 2023-08-04
XmlNode/XmlElement as an Input Parameter

bfcrmimick:
I have recently experienced the Null value described in the issue https://github.com/DigDes/SoapCore/issues/908 and decided to see if I could follow the suggested changes.  In this pull request, I have added the suggested fix from @andersjonsson and then I also added a few Unit Tests for it.  

In the integration tests, we have an XmlElement return type only and then a second test with both an XmlElement input and an XmlElement return time. 

I have also added a new folder called XmlNodeInputOutputTests for the purpose of testing the XmlElement functions as a SOAP function vs a ChannelFactory object.  I found the code is not quite happy with an XmlNode return type (the Test Server returns a message stating that it didn't get enough to consider a valid response).  

Thank you,

Raymond Mimick

### DIFF
--- a/src/SoapCore.Tests/ITestService.cs
+++ b/src/SoapCore.Tests/ITestService.cs
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using System.ServiceModel;
 using System.Threading.Tasks;
+using System.Xml;
 using Microsoft.AspNetCore.Mvc;
 using SoapCore.Tests.Model;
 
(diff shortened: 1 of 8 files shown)


bfcrmimick:
Apologies for the failing test case - it was a matter of Windows vs Linux in running the test when replacing new line characters.  Updated to use Environment.NewLine instead.

andersjonsson:
Looks good. Is there anything else you would like to add before merging? I think it's acceptable if it doesn't work with an XmlNode return type, since it's not a regression.

bfcrmimick:
At the moment, I’m ok.  I was able to successfully test in my own development project with the changes so that worked better.  Thank you for the descriptions you added in the issue to help us. 

 

Raymond Mimick

andersjonsson:
Thank you for your contribution

PR #965, opened 2023-09-26
Added support for array of arrays parameters in methods and a test

flipm0de:


flipm0de:
The test and description should be self explanatory, but in some cases we have SOAP methods which accept parameters with types like byte [][]. In such cases we get an XmlException in the middleware. With this change we get the correct behavior.

PR #974, opened 2023-10-27
Update SoapEndpointMiddleware.cs with Request.PathBase

JohnnyAtHogia:
To generate correct ServerUrl when hosting the soap site under a existing site as an application you would need to also have the Request.PathBase within the ServerUrl generation.

Example hosting a SoapCore site under default website locally (http://localhost) in a application with path trunk/sopcoresite without the Request.PathBase the location in wsdl is generated incorrectly.

http://localhost/trunk/soapcoresite/checkin.asmx?wsdl 
location="http://localhost/checkin.asmx"

With Request.PathBase it will set the correct location. 
location="http://localhost/trunk/soapcoresite/checkin.asmx"

localhost_trunk_soapcoresite_checkin asmx_wsdl
localhost_trunk_soapcoresite_checkin asmx_wsdl-PathBase

### DIFF
--- a/src/SoapCore/SoapEndpointMiddleware.cs
+++ b/src/SoapCore/SoapEndpointMiddleware.cs
@@ -983,7 +983,7 @@ namespace SoapCore
 			}
 			else
 			{
-				meta.ServerUrl = httpContext.Request.Scheme + "://" + httpContext.Request.Host + "/";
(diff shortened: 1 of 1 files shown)


JohnnyAtHogia:
Learned that virtual path does the same thing and if this is accepted it will be a breaking change for those that already uses virtual path as intended.

PR #986, opened 2024-01-07
support for XmlEnumAttribute specifiying custom names for enum members

edgridin:
Feature request: support for XmlEnumAttribute to be able to specifiy custom names for enum members

for the example enum, used in test CheckEnumWithCustomNamesXmlSerializedWsdl

```
public enum EnumWithCustomNames
	{
		[XmlEnum("F")]
		FirstEnumMember,

		[XmlEnum("S")]
		SecondEnumMember,

		ThirdEnumMember
	}

```
the expected output would be

```
<xsd:simpleType name="EnumWithCustomNames">
        <xsd:restriction base="xsd:string">
          <xsd:enumeration value="F" />
          <xsd:enumeration value="S" />
          <xsd:enumeration value="ThirdEnumMember" />
        </xsd:restriction>
      </xsd:simpleType>
```

### DIFF
--- a/src/SoapCore/Meta/MetaBodyWriter.cs
+++ b/src/SoapCore/Meta/MetaBodyWriter.cs
@@ -462,10 +462,27 @@ namespace SoapCore.Meta
 					writer.WriteStartElement("restriction", Namespaces.XMLNS_XSD);
 					writer.WriteAttributeString("base", $"{_xmlNamespaceManager.LookupPrefix(Namespaces.XMLNS_XSD)}:string");
 
-					foreach (var value in Enum.GetValues(toBuild))
+					var membersWithCustomNames = toBuild.GetMembersWithAttribute<XmlEnumAttribute>();
+					foreach (var value in Enum.GetNames(toBuild))
 					{
+						var xmlEnumAttribute = membersWithCustomNames
+												.FirstOrDefault(m => m.Name == value)?
+												.CustomAttributes
+												.FirstOrDefault(c => c.AttributeType == typeof(XmlEnumAttribute));
+						var nameFromAttribute = (xmlEnumAttribute?
+													.ConstructorArguments?
+													.FirstOrDefault()
+													.Value
+												??
+												xmlEnumAttribute?
+													.NamedArguments?
+													.FirstOrDefault(a => a.MemberName == "Name")
+													.TypedValue
+													.Value)?
+												.ToString();
+
 						writer.WriteStartElement("enumeration", Namespaces.XMLNS_XSD);
-						writer.WriteAttributeString("value", value.ToString());
+						writer.WriteAttributeString("value", nameFromAttribute ?? value);
 						writer.WriteEndElement(); // enumeration
 					}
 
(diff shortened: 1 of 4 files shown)


src/SoapCore/Meta/MetaBodyWriter.cs
```diff
@@ -462,10 +462,27 @@ private void AddTypes(XmlDictionaryWriter writer)
...
+													.FirstOrDefault()
+													.Value
+												??
+												xmlEnumAttribute?
+													.NamedArguments?
+													.FirstOrDefault(a => a.MemberName == "Name")
+													.TypedValue
+													.Value)?
+												.ToString();
+
```
andersjonsson: This code traverses all members with custom names for every member resulting in a potential O(n^2) perf characteristic (since it traverses all members with custom attribute once per member in the enum)

Please see if you can improve that. I gave it a shot, and my approach is included below. Feel free to use that implementation if you want to.

I tried optimizing the code with an approach like this
```
Type toBuild = typeof(MyEnum);
var membersWithCustomNames = from n in Enum.GetNames(toBuild)
							 join m in toBuild.GetMembers() on n equals m.Name
							 let ca = m.CustomAttributes.FirstOrDefault(ca => ca.AttributeType == typeof(XmlEnumAttribute))
							 select new
							 {
								 Name = m.Name,
								 CustomName =
								 (ca?
												.ConstructorArguments?
												.FirstOrDefault()
												.Value
											??
											ca?
												.NamedArguments?
												.FirstOrDefault(a => a.MemberName == "Name")
												.TypedValue
												.Value)?
											.ToString()
							 };
foreach (var value in membersWithCustomNames)
{
	var finalValue = value.CustomName ?? value.Name;
}
```

Results from BenchmarkDotnet (my implementation being the second one). I used an Enum with 10 members, 4 of which had XmlEnum-attributes.
image

andersjonsson (changes requested):
Thank you very much for the PR!

Looks good! I left a comment regarding the performance. Please take a look at that and we can get this PR merged shortly

edgridin:
hi Anders,

wow, I didn't expect, that a Linq query with a join would have better performance, than just looping and querying for single members. But yeah, the difference is huge, so I've changed the implementation to your proposal.

Maybe an additional question - I've seen, that wsdl generation with XmlSerializer currently doesn't support SOAP fault contracts. Do you know, if someone is already on it? Otherwise I could try to implement it.

Best regards,
Eduard

andersjonsson:
> hi Anders,
> 
> wow, I didn't expect, that a Linq query with a join would have better performance, than just looping and querying for single members. But yeah, the difference is huge, so I've changed the implementation to your proposal.

Yeah, it's the query per member that's killing perf. That same "pattern" has caused me lots of headache over the years 😅 
A pre-constructed dictionary, so that you wouldn't have to run a linq-query per iteration, would work just as well.

> Maybe an additional question - I've seen, that wsdl generation with XmlSerializer currently doesn't support SOAP fault contracts. Do you know, if someone is already on it? Otherwise I could try to implement it.

Not to my knowledge. You are more than welcome to give it a go 😄

andersjonsson:
Merged! Thanks again!

edgridin:
> > Maybe an additional question - I've seen, that wsdl generation with XmlSerializer currently doesn't support SOAP fault contracts. Do you know, if someone is already on it? Otherwise I could try to implement it.
> 
> Not to my knowledge. You are more than welcome to give it a go 😄

OK, I'll give it a try :-)

Could you maybe publish the new package version? This XmlEnum stuff is really needed, otherwise I'd have an enum with one-letter-members in my current project, which is terrible 😄

andersjonsson:
Done. A new version should be available on nuget in a couple of minutes

PR #991, opened 2024-01-19
fix: correct wsdl json codeblock in readme

rohanorton:
The JSON codeblock in the README was invalid. I have made the following changes:

- Removed syntax error (floating comma)
- Added missing curly brace
- Changed block syntax from `csharp` to `json`
- Reformatted


andersjonsson:
Thanks

PR #995, opened 2024-01-30
Fix bug ArrayOfString maxOccurs="1" bug

### LINKED ISSUE #988: ArrayOfString maxOccurs 1
Hi!

```html
<xsd:complexType name="ArrayOfString">
    <xsd:sequence>
        <xsd:element minOccurs="0" maxOccurs="1" name="string" type="xsd:string"/>
    </xsd:sequence>
</xsd:complexType>
```
For unknown reason when I use SoapCore `ArrayOfString` in wsdl generated with `maxOccurs="1"` instead of `maxOccurs="unbounded"` attribute like it generates in WCF.
I'm using xmlSerializer

What info I could provide?

vidrenning:
This is a regression of #960.
In commit 5026da3f41660ab732b6aa5491a17d0aef33c17e, the AddSchemaType() method was changed to contain a special case for string and byte[]. However this breaks the situation where the inner type of a list is string. In that case, maxOccurs must be set to "unbounded" for the string element. So this commit reverts that change and instead fixes the original issue at another place.

Fixes #988

### DIFF
--- a/src/SoapCore.Tests/Wsdl/Services/ComplexType.cs
+++ b/src/SoapCore.Tests/Wsdl/Services/ComplexType.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Xml.Serialization;
 
 namespace SoapCore.Tests.Wsdl.Services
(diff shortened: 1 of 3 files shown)


andersjonsson:
Nice! Thank you!

vidrenning:
Thank you for merging the change!

PR #997, opened 2024-01-31
WSDL for XmlSerializer: Fix invalid ArrayOfString schema

### LINKED ISSUE #996: XmlSerializer: Invalid namespace in WSDL for List<string> type
The generated WSDL for ArrayOfString objects is wrong and does not correspond to the XML that XmlSerializer actually generates when calling the web service.

```csharp
public class HelloResult
{
    public List<string> StringList { get; set; } = [];
}

[ServiceContract]
public class HelloService
{
    [OperationContract]
    public HelloResult Hello()
    {
        return new HelloResult { StringList = [ "test" ] };
    }
}
```

The following WSDL will get generated:
```xml
<xsd:complexType name="HelloResult">
  <xsd:sequence>
    <xsd:element minOccurs="0" maxOccurs="1" xmlns:q1="http://schemas.microsoft.com/2003/10/Serialization/Arrays" name="StringList" nillable="true" type="q1:ArrayOfString"/>
  </xsd:sequence>
</xsd:complexType>
```
In the generated WSDL, ArrayOfString is in the namespace http://schemas.microsoft.com/2003/10/Serialization/Arrays and is further below defined
```xml
<xsd:schema xmlns:tns="http://schemas.microsoft.com/2003/10/Serialization/Arrays" elementFormDefault="qualified" targetNamespace="http://schemas.microsoft.com/2003/10/Serialization/Arrays">
  <xsd:complexType name="ArrayOfString">
    <xsd:sequence>
      <xsd:element minOccurs="0" maxOccurs="1" name="string" type="xsd:string"/>
    </xsd:sequence>
  </xsd:complexType>
  <xsd:element name="ArrayOfString" nillable="true" type="tns:ArrayOfString"/>
</xsd:schema>
```
Ignore the fact that macOccurs is set to 1 here, this has already been reported in #988. The actual issue is that the XML that is actually transferred when calling the web service is different. I have used Wireshark to check the actual SOAP response that is returned from XmlSerializer:
```xml
<s:Envelope xmlns:s="http://www.w3.org/2003/05/soap-envelope" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <s:Body>
    <HelloResponse xmlns="http://tempuri.org/">
      <HelloResult>
        <StringList>
          <string>test</string>
        </StringList>
      </HelloResult>
    </HelloResponse>
  </s:Body>
</s:Envelope>
```

As you can see here, the XmlSerializer will not use the http://schemas.microsoft.com/2003/10/Serialization/Arrays namespace at all. Only the DataContractSerializer will use http://schemas.microsoft.com/2003/10/Serialization/Arrays. The generated WSDL is wrong for XmlSerializer and should be fixed.

vidrenning:
Removed the special if-condition for "string", which does not make sense, because it does not correspond to the actual XML that is generated by the XmlSerializer. Afterwards, the field _arrayToBuild is not needed anymore, so I have removed it.

The test CheckStringArrayNameWsdl() can only be performed for the DataContractSerializer now, because the complexType does not have an element inside it anymore.

Fixes #996

### DIFF
--- a/src/SoapCore.Tests/Wsdl/WsdlTests.cs
+++ b/src/SoapCore.Tests/Wsdl/WsdlTests.cs
@@ -481,7 +481,6 @@ namespace SoapCore.Tests.Wsdl
 		}
 
 		[DataTestMethod]
-		[DataRow(SoapSerializer.XmlSerializer)]
 		[DataRow(SoapSerializer.DataContractSerializer)]
 		public async Task CheckStringArrayNameWsdl(SoapSerializer soapSerializer)
 		{
--- a/src/SoapCore/Meta/MetaBodyWriter.cs
+++ b/src/SoapCore/Meta/MetaBodyWriter.cs
@@ -28,11 +28,9 @@ namespace SoapCore.Meta
 
 		private readonly Queue<Type> _enumToBuild;
 		private readonly Queue<TypeToBuild> _complexTypeToBuild;
-		private readonly Queue<Type> _arrayToBuild;
 
 		private readonly HashSet<string> _builtEnumTypes;
 		private readonly HashSet<string> _builtComplexTypes;
-		private readonly HashSet<string> _buildArrayTypes;
 		private readonly Dictionary<string, Dictionary<string, string>> _requestedDynamicTypes;
 
 		private bool _buildDateTimeOffset;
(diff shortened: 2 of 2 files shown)


andersjonsson:
Thank you!

vidrenning:
Thank you for merging!

PR #1003, opened 2024-02-05
Installing vulnerable nuget packages into project explicitly

programmersommer:
Some nuget packages have some not safe transitive dependencies

image

Newer versions of those nuget packages have been installed into project


andersjonsson:
Thanks!

PR #1007, opened 2024-02-07
XmlSerializer WSDL: XmlAttribute: Set use="required" attribute

### LINKED ISSUE #1006: XmlSerializer: WSDL for XmlAttribute fields does not set use="required" flag
The WSDL schema that is generated for fields or properties with the `[XmlAttribute]` attribute is not correct. It differs from the WSDL that is generated by a ASP.NET ASMX Web Service. Let's take a look at the following class:
```csharp
public class AttributeType
{
    [XmlAttribute]
    public int IntProperty { get; set; }

    [XmlAttribute]
    public int OptionalIntProperty { get; set; }

    public bool ShouldSerializeOptionalIntProperty()
    {
        return OptionalIntProperty != 0;
    }
}
```

SoapCore will generate a WSDL, such as the following:
```xml
<xsd:attribute name="IntProperty" type="xsd:int" />
<xsd:attribute name="OptionalIntProperty" type="xsd:int" />
```
However the WSDL that is generated by ASMX Web Services in .NET Framework is different. It adds the use="required" attribute for value types (such as int), to indicate that they are required. The required attribute is added for all value types, except if the class defines a `ShouldSerialize*()` method, which makes the type optional again.

This is the how the generated WSDL should actually look:
```xml
<xsd:attribute name="IntProperty" type="xsd:int" use="required" />
<xsd:attribute name="OptionalIntProperty" type="xsd:int" />
```

This applies to the current version v1.1.0.42.

vidrenning:
If a member has been annotated with the [XmlAttribute] attribute, and it is a value type, such as int, it will get the use="required" attribute. However if it has a ShouldSerialize*() method, then it is considered optional and the use="required" attribute is not added.

Added a new test to verify this behavior.

Fixes #1006

### DIFF
--- a/src/SoapCore.Tests/Wsdl/Services/AttributeType.cs
+++ b/src/SoapCore.Tests/Wsdl/Services/AttributeType.cs
@@ -1,0 +1,21 @@
+using System.Xml.Serialization;
+
+namespace SoapCore.Tests.Wsdl.Services
+{
+	public class AttributeType
+	{
+		[XmlAttribute]
+		public string StringProperty { get; set; }
+
+		[XmlAttribute]
+		public int IntProperty { get; set; }
+
+		[XmlAttribute]
+		public int OptionalIntProperty { get; set; }
+
+		public bool ShouldSerializeOptionalIntProperty()
+		{
+			return OptionalIntProperty != 0;
+		}
+	}
+}
--- a/src/SoapCore.Tests/Wsdl/Services/IAttributeService.cs
+++ b/src/SoapCore.Tests/Wsdl/Services/IAttributeService.cs
@@ -1,0 +1,24 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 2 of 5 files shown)


andersjonsson (approved):
Looks good! 

Maybe you should check that the return type of the ShouldSerialize*()-methods is System.Boolean.
On the other hand serialization will fail if the return type is something else, so the user that does something like that will have bigger issues than faulty wsdl 😛

vidrenning:
Thanks for merging! Before implementing this I have tested the behavior of the .NET Framework ASMX Web Service WSDL generator. In that generator, the return type of the `ShouldSerialize*()` methods is also not checked. The WSDL is the same, no matter if a `int ShouldSerialize*()` method or a `bool ShouldSerialize*()` method is defined. Therefore I decided to also not check it, to maximize compatibility with the .NET Framework ASMX Web Service WSDL generator. 😄

PR #1009, opened 2024-02-07
XmlSerializer WSDL: Use uppercase letter for ArrayOf<Type> type

### LINKED ISSUE #1008: XmlSerializer WSDL: Generated type ArrayOf<Type> should always use uppercase first letter
Let's say we have a type such as
```csharp
public class ArrayType
{
    public List<innerType> InnerTypeList { get; set; } = [];
}

public class innerType
{
    public string Name { get; set; } = "";
}
```

As you can see the name of the class `innerClass` starts with a lowercase "i".

For this class, SoapCore will generate a type with the name `ArrayOfinnerClass`. However the WSDL that is generated by a ASMX web service in .NET Framework is different. It will generate a type with the name `ArrayOfInnerClass`. As you can see, it makes the first letter uppercase.

vidrenning:
The first letter of the class name is now converted to upper-case.

Added a test that verifies the correct behavior.

Fixes #1008


andersjonsson:
Seems reasonable

vidrenning:
Thanks for merging!

PR #1010, opened 2024-02-07
Add option IndentWsdl to control if the generated WSDL should be indented

vidrenning:
Currently the option `IndentXml` can be used to control if the generated XML is indented or not. This applies to both: the SOAP response XML and the generated WSDL.

It would be nice to control both of these independently. For example, the XML for the SOAP responses should not be indented, but the XML for the WSDL should be indented.

I propose to add a new configuration option `IndentWsdl`. If this option is explicitly set, it will control if the generated WSDL is indented or not. If it is not set, it will default to the setting of `IndentXml`, so existing users should not notice a difference in behavior.


andersjonsson:
Looks good. Thanks!

vidrenning:
Thanks for merging!

PR #1012, opened 2024-02-07
XmlSerializer WSDL: Improve enum order

### LINKED ISSUE #1011: XmlSerializer WSDL: Enum order is different than in ASMX Web Service
Lets say there is a enum such as

```csharp
public enum EnumWithCustomNames
{
    FirstEnumMember = -2,
    SecondEnumMember = 1,
    ThirdEnumMember = 0
}
```

SoapCore will generate the following WSDL:
```xml
<xsd:restriction base="xsd:string">
  <xsd:enumeration value="ThirdEnumMember" />
  <xsd:enumeration value="SecondEnumMember" />
  <xsd:enumeration value="FirstEnumMember" />
</xsd:restriction>
```

However an ASXM Web Service in .NET Framework will generate the following WSDL:

```xml
<xsd:restriction base="xsd:string">
  <xsd:enumeration value="FirstEnumMember" />
  <xsd:enumeration value="SecondEnumMember" />
  <xsd:enumeration value="ThirdEnumMember" />
</xsd:restriction>
```

The WSDL that is generated by the .NET Framework ASMX Web Service will always specify the enum values in the order that they are declared in the source code. However SoapCore will specify them in a different order: First zero, then all positive values, then all negative values (order of the `Enum.GetNames()` method)

To maximize compatibility, I think that the enum value order should be changed so that it is the source code order.

vidrenning:
The enum values are now specified in the order in which they are declared in the source code. This is the same behavior as the WSDL that is generated by a .NET Framework ASMX Web Service.

Improved the enum test so that it also checks the order of the elements.

Fixes #1011

### DIFF
--- a/src/SoapCore.Tests/Wsdl/Services/EnumWithCustomNames.cs
+++ b/src/SoapCore.Tests/Wsdl/Services/EnumWithCustomNames.cs
@@ -5,11 +5,11 @@ namespace SoapCore.Tests.Wsdl.Services
 	public enum EnumWithCustomNames
 	{
 		[XmlEnum("F")]
-		FirstEnumMember,
+		FirstEnumMember = -2,
 
 		[XmlEnum("S")]
-		SecondEnumMember,
+		SecondEnumMember = 1,
 
-		ThirdEnumMember
+		ThirdEnumMember = 0
 	}
 }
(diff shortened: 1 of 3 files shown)


andersjonsson:
Nice!

vidrenning:
Thanks for merging!

PR #1016, opened 2024-02-09
Do not use chunked transfer encoding on ASP.NET Core 2.1

### LINKED ISSUE #1015: SoapCore uses chunked transfer-encoding on netstandard2.0 version
When using SoapCore with .NET Framework and ASP.NET Core 2.1, chunked transfer encoding is used. I have run into some compatibility issues with some existing applications, so I think chunked transfer-encoding should not be used. This has already been implemented in #843 for ASP.NET Core >= 3.0, but on ASP.NET Core 2.1, chunked transfer-encoding is still used.

vidrenning:
Setting the Content-Length header explicitly will disable chunked transfer-encoding. However this was only done for one of the two WriteMessageAsync methods in SoapMessageEncoder. I have changed the other method to also set the Content-Length header.

To do this I have used a MemoryStream. The other implementation used a StringWriter, which is not ideal, because it will convert to and from different encodings. We only want to find out the length of the stream, so a MemoryStream is faster and sufficient. Also changed the other WriteMessageAsync implementation to use a MemoryStream and removed the CustomStringWriter class.

Fixes #1015


vidrenning:
Thanks for merging!

PR #1017, opened 2024-02-12
Add option to generate soapAction without contract name

vidrenning:
By default, the soapAction that is generated if it is not specified in the `[OperationContract(Action = "...")]` attribute, follows the format `{namespace}/{contractName}/{methodName}`.

This seems to be different than the default soapAction that is generated in ASMX web services in .NET Framework. The new option makes it possible to omit the contractName and instead generate soapActions with the format `{namespace}/{methodName}`. The default for this new option is false. This means that the behavior for existing users of SoapCore does not change.

However I had to change the interface ISoapMessageProcessor to make AuthorizeOperationMessageProcessor work. The method ProcessMessage will now have an additional SoapOptions parameter. If users implement their own ISoapMessageProcessor, they have to add this parameter if they update the NuGet package. Is this acceptable?

### DIFF
--- a/src/SoapCore.Tests/OperationDescription/OperationDescriptionTests.cs
+++ b/src/SoapCore.Tests/OperationDescription/OperationDescriptionTests.cs
@@ -11,14 +11,14 @@ namespace SoapCore.Tests
 		[Fact]
 		public void TestProperUnrappingOfGenericResponses()
 		{
-			ServiceDescription serviceDescription = new ServiceDescription(typeof(IServiceWithMessageContract));
-			ContractDescription contractDescription = new ContractDescription(serviceDescription, typeof(IServiceWithMessageContract), new ServiceContractAttribute());
+			ServiceDescription serviceDescription = new ServiceDescription(typeof(IServiceWithMessageContract), false);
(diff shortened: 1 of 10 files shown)


vidrenning:
After thinking some more about it, I have realized that it is not necessary to change the `ISoapMessageProcessor` interface. I have updated the commit to reflect that. So users do not need to add a new parameter when updating the NuGet package.

andersjonsson:
Great! Easier if we can avoid breaking changes :)
I'll try to take a look at the PR today

andersjonsson:
Thanks!

vidrenning:
Thank you very much for merging! Sorry to bother you again with this, but would it be possible to build a new NuGet package version again? I have evaluated that no other changes are required for my situation and having a new NuGet package would mean that I can go through with the deployment 🙂.

vidrenning:
Thank you! 🙂

andersjonsson:
> Thank you! 🙂

Np. Sorry that I didn't let you know about the new version immediately. A little bit too much going on at work right now and my work on SoapCore is a spare time thing.

PR #1019, opened 2024-02-19
support wsdl:import and simplify the node name comparsions (prefix:loca…

zgabi:
…lnames)

### DIFF
--- a/src/SoapCore.Tests/SoapCore.Tests.csproj
+++ b/src/SoapCore.Tests/SoapCore.Tests.csproj
@@ -61,6 +61,18 @@
 	</ItemGroup>
 
 	<ItemGroup>
+		<None Update="WsdlFromFile\WSDLInclude\echoWsdlInclude.wsdl">
+		  <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+		</None>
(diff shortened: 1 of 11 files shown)


andersjonsson:
Please fix the failing test

zgabi:
Ok, please check it now.. Sorry for the bug, I forget to fix 1 thing. Now it should be OK.

andersjonsson:
There are also issues in the Code scanning. It doesn't like trusting Xml from an external source

There seems to be more info, and a suggested fix, here
https://github.com/dotnet/roslyn-analyzers/issues/2477

zgabi:
This insecure thing is not a new problem. The affected line for XSD was already there:
https://github.com/DigDes/SoapCore/blob/ec9b94ce30fb89d4b8c097890e2847f13c77218d/src/SoapCore/SoapEndpointMiddleware.cs#L989

Should I also fix that problem?

zgabi:
Maybe it is caused by the rename (xsdfile  => xsdFile)... but in this case it was already there just the CodeQL did not noticed it :)

zgabi:
I added the suggested fix, please check it again.
I also added unit test for the WSDL import.

andersjonsson:
Thanks!

zgabi:
Thank you, too

PR #1023, opened 2024-02-26
XmlSerializer: Remove empty <xsd:import /> tag in generated WSDL

### LINKED ISSUE #1022: XmlSerializer: Empty `<xsd:import />` tag in generated WSDL
This is an issue that got introduced by my commit for #1014. Sorry about that!

In that commit I wanted to remove the line `<xsd:import namespace="http://schemas.datacontract.org/2004/07/System" />` from the generated WSDL. However in the actual commit, I only removed the `namespace="http://schemas.datacontract.org/2004/07/System"` part, so an empty `<xsd:import />` tag remains in the WSDL.

It turns out that this throws off some third party SOAP parsers, so it is better to remove it.

vidrenning:
Fixes #1022


andersjonsson:
Thanks!

vidrenning:
Thanks for merging! 🙂

PR #1024, opened 2024-02-26
Add Option to not normalize newlines

vidrenning:
The XmlWriter that is used by SoapCore to write the SOAP XML responses for all requests will by default normalize newlines in the generated XML responses.

This means that for a string that contains LF (`'\n'`) newlines, these newlines will automatically be converted to CRLF (`"\r\n"`).

This is different from the behavior of the old ASMX Web Services, which will not normalize the newlines.

I have added a new option that controls whether the newlines should be normalized or not. The default behavior is "true", so the commit will not change the behavior for existing users of SoapCore.


andersjonsson:
Seems like a good thing. Thanks!

vidrenning:
Thanks for merging! Would it be possible to build a new NuGet version again? Thank you for all your work on maintaining SoapCore! 🙂

andersjonsson:
Nuget package on the way. 

The new package also includes a change to avoid blocking when the request-message is read. Reading will now take longer and use more memory but hopefully the overall throughput is increased. If you notice performance regressions please submit an issue

vidrenning:
Thank you, appreciate it! I will let you know if I notice any regressions.

Issue #1021, opened 2024-02-24
500 Error code for a Bad XML body 


akshaybheda:
I was trying to get my service out to production, it gets 100 req per second per instance, i am replacing the old WCF with SOAPCore, I started to see these error. I tried to figure out whats going on but I don't log req headers or body in the logs, 

this is the stacktrace i found for the exception

```
System.Xml.XmlException:
   at System.Xml.XmlExceptionHelper.ThrowXmlException (System.Private.DataContractSerialization, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.Xml.XmlUTF8TextReader.Read (System.Private.DataContractSerialization, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.Xml.XmlBaseReader.ReadEndElement (System.Private.DataContractSerialization, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.ServiceModel.Channels.Message.ReadFromBodyContentsToEnd (System.Private.ServiceModel, Version=4.8.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.ServiceModel.Channels.ReceivedMessage.OnWriteBodyContents (System.Private.ServiceModel, Version=4.8.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.ServiceModel.Channels.Message.OnWriteMessage (System.Private.ServiceModel, Version=4.8.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.ServiceModel.Channels.Message.OnCreateBufferedCopy (System.Private.ServiceModel, Version=4.8.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at SoapCore.HeadersHelper.GetSoapAction (SoapCore, Version=1.1.0.41, Culture=neutral, PublicKeyToken=092c4b76488e2b26: /_/src/SoapCore/HeadersHelper.cs:21)
   at SoapCore.SoapEndpointMiddleware`1+<ProcessMessage>d__17.MoveNext (SoapCore, Version=1.1.0.41, Culture=neutral, PublicKeyToken=092c4b76488e2b26: /_/src/SoapCore/SoapEndpointMiddleware.cs:455)
[...]

PR #1025, opened 2024-02-26
added bad request status code for Ill formed xml requests: XML Except…

akshaybheda:
Currently if there is an incorrect character at the end of the request body after > tag. It throws internal server error where it should be bad request


akshaybheda:
Fixes #1021

andersjonsson:
Seems reasonable

andersjonsson:
Thanks!

PR #1028, opened 2024-02-28
updated csproj version

akshaybheda:


akshaybheda:
@andersjonsson can you re-run nuget pulblish action?

PR #1029, opened 2024-02-29
removed unnessecary buffered copy

andersjonsson:
* added method to deserialize SoapDefinition from stream 
* removed unnessecary StreamReader usage

@zgabi made me aware that MemoryStream does not need to be disposed.
That made me realize that we could avoid a BufferedCopy of message in ReadMessageAsync.

Then I found some other perf fixes that could be done around memorystream usage in the lib.

@akshaybheda all that hoop jumping and this code is back to basically your first commit in ReadMessageAsync. 😆

### DIFF
--- a/src/SoapCore/DocumentationWriter/SoapMethods.cs
+++ b/src/SoapCore/DocumentationWriter/SoapMethods.cs
@@ -49,6 +49,15 @@ namespace SoapCore.DocumentationWriter
 			return (SoapDefinition)xs.Deserialize(sr);
 		}
 
+		public static SoapDefinition DeserializeFromStream(Stream xml)
+		{
+			XmlSerializer xs = new XmlSerializer(typeof(SoapDefinition));
+
+			xs.UnknownElement += _unknownElementHandler;
+
+			return (SoapDefinition)xs.Deserialize(xml);
+		}
+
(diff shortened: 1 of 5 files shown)


akshaybheda (approved):
these actually look good, even i wasn't are about disposing of memory stream is not required

akshaybheda:
Did you run benchmark for this?

andersjonsson:
> Did you run benchmark for this?

I did. Different payload from last time, but this looks good

```
| Method                           | Mean     | Error     | StdDev    | Median   | Gen0   | Allocated |
|--------------------------------- |---------:|----------:|----------:|---------:|-------:|----------:|
| ReadMessageAsyncWithMemoryStream | 6.147 μs | 0.1226 μs | 0.3053 μs | 6.040 μs | 4.7455 |  19.39 KB |
| ReadMessageAsync                 | 5.342 μs | 0.1060 μs | 0.1856 μs | 5.333 μs | 3.4103 |  13.93 KB |
```

Just using a MemoryStream is obviously much cheaper than a buffered copy of the Message

akshaybheda (approved):

akshaybheda:
looking forward to new version, also appreciate  your time on this

PR #1031, opened 2024-02-29
Dotnet8

andersjonsson:
also updated nuget packages


src/SoapCore/SoapCore.csproj
```diff
@@ -2,9 +2,9 @@
 
 	<PropertyGroup>
 		<Description>SOAP protocol middleware for ASP.NET Core</Description>
-		<Version>1.1.0.49</Version>
+		<Version>1.2.0.0</Version>
 		<Authors>Digital Design</Authors>
-		<TargetFrameworks>netstandard2.0;netstandard2.1;netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
+		<TargetFrameworks>netstandard2.0;netstandard2.1;net7.0;net8.0</TargetFrameworks>
```
thompson-tomo: I don't see the benefit in removing net core 3.1, net 5/6. At the same time only marginal benefit of having net 7. I would propose we first merge in #1052

thompson-tomo (changes requested):

src/SoapCore/SoapCore.csproj
andersjonsson: I haven't had the time to work on this one for a while. Will probably resume work in a couple of weeks.
net core 3.1 and net 5 are already covered by targeting netstandard.
Agree about net 7 though. It's already out of support and I don't see a reason to keep it around

thompson-tomo: Looking at the dependency graph if you want to reduce target framework I would stick with net core 3 as otherwise u would be forcing all projects < net 8 to take on a deprecated dependencies.

noam-amiran: does this works for you? we had some difficulties making this work

HugoVG: When will this be merged with the already merged pr's with a new nuget package?

akshaybheda:
Had a question, any tentative timeline for the release?

cime:
I've merged the origin/develop into @andersjonsson dotnet8 branch, and did some other changes.
Please take a look and feel free to take any changes back to this PR or I can open a new PR.

https://github.com/cime/SoapCore/tree/dotnet8

akshaybheda30:
@andersjonsson Net 6 goes out of support in Nov, could we expect a release in oct for this?

andersjonsson:
> @andersjonsson Net 6 goes out of support in Nov, could we expect a release in oct for this?

Sorry for the lack of activity on this branch. Life...
@cime Thanks for continuing the work! Please open a new PR and I'll take a look

Thompson1985 (approved):

andersjonsson:
I've merged origin/develop into andersjonsson/dotnet8
I removed net7 support but kept netcoreapp3.1 around, for better or worse

akshaybheda (approved):

PR #1034, opened 2024-03-05
support for X-Forwarded-Host header

zgabi:
The X-Forwarded-Host is a de-facto standard header which is set by the reverse proxies.
This PR adds support for this header. It also adds supoprt to override the url during the wsql/xsd query (with inheriting from the WsdlFileOptions class) and allows to override the http scheme.

For example it is useful in the following configuration:

The service is accessible from the internet in the following url:
https://external.hostname.com/ServiceName

But it is only a proxy, calls the internal service:
http://internal.hostname.com/ServiceName

And we also have development and test urls:
https://test.hostname.com/ServiceName
http://internaltest.hostname.com/ServiceName

The proxy sets the X-Forwarded-Host header (external.hostname.com or test.hostname.com), and SoapCore overrides the http scheme to "http".

In this way the address in the wsdl will be correct even if it is called from the internet or from the internal url directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ To use it, add a setting like this to appsettings
 ```json
 "FileWSDL": {
   "UrlOverride": "",
+  "SchemeOverride": "",
   "VirtualPath": "",
   "WebServiceWSDLMapping": {
     "Service.asmx": {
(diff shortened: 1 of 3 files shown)


andersjonsson:
Would it be easier to add a `Func<HttpContext, string> UrlOverrideFunc` that one can set instead of using inheritance?
Either way is fine by me, I just wanted to put the option out there

zgabi:
Ok, I'll modify it today evening or tomorrow (I can't do it right now) Please do not merge it until than.

andersjonsson:
Oh god damn it... I'm too tired and accidently merged the wrong PR. I'm sorry... 😞

andersjonsson:
Reverted the PR but I'm afraid you'll have to resubmit a PR

zgabi:
Sorry, my yesterday was very busy... I do it now.. sumbit a new PR in some hours... thanks.

PR #1035, opened 2024-03-05
the com exceptions are just bad requests which should not be 500

akshaybheda:
This is to prevent https://github.com/dotnet/aspnetcore/issues/45086

### DIFF
--- a/src/SoapCore/SoapEndpointMiddleware.cs
+++ b/src/SoapCore/SoapEndpointMiddleware.cs
@@ -14,7 +14,6 @@ using System.ServiceModel.Channels;
 using System.Text;
 using System.Threading.Tasks;
 using System.Xml;
-using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Mvc;
@@ -326,7 +325,7 @@ namespace SoapCore
 				{
 					status = StatusCodes.Status400BadRequest;
 				}
-				else if ((ex is ConnectionResetException) && (ex?.Message == "The client has disconnected"))
+				else if (ex?.Message == "The client has disconnected")
 				{
 					status = StatusCodes.Status400BadRequest;
 				}
--- a/src/SoapCore/SoapEndpointMiddleware.cs
+++ b/src/SoapCore/SoapEndpointMiddleware.cs
@@ -7,6 +7,7 @@ using System.Net;
 using System.Net.Http.Headers;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Security.Authentication;
 using System.ServiceModel;
 using System.ServiceModel.Channels;
@@ -321,6 +322,10 @@ namespace SoapCore
 					status = StatusCodes.Status403Forbidden;
 				}
 				else if (ex is XmlException)
+				{
+					status = StatusCodes.Status400BadRequest;
+				}
+				else if ((ex is COMException) && (ex?.Message == "The client has disconnected"))
 				{
 					status = StatusCodes.Status400BadRequest;
 				}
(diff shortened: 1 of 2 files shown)


src/SoapCore/SoapEndpointMiddleware.cs
```diff
@@ -324,6 +325,10 @@ private async Task ProcessOperation(HttpContext httpContext, IServiceProvider se
 				{
 					status = StatusCodes.Status400BadRequest;
 				}
+				else if (ex?.Message == "The client has disconnected")
+				{
+					status = StatusCodes.Status400BadRequest;
+				}
```
andersjonsson: Error messages can change so it's best not to compare against those. 

```suggestion
				else if (ex is ConnectionResetException)
				{
					status = StatusCodes.Status400BadRequest;
				}

```

Using ConnectionResetException requires that we reference an additional nuget package but that's fine.

If you only want to check for the "The specified network name is no longer available"-exception you can do something like this:

`if(ex?.InnerException is COMException cex && cex.ErrorCode == unchecked((int)0x80070040))`

A more radical option would be to go even wider and just check for IOException. That is probably fine also

andersjonsson (changes requested):

src/SoapCore/SoapEndpointMiddleware.cs
akshaybheda: I originally went with ConnectionResetException  but i am not sure which version to install since this supports multiple frameworks

akshaybheda: This is the issue 
image

andersjonsson: This works for me (manually edit the .csproj file)
```
<ItemGroup Condition="$(TargetFramework) == 'netstandard2.0' OR $(TargetFramework) == 'netstandard2.1'">
	<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
	<PackageReference Include="Microsoft.AspNetCore.Http" Version="2.1.34" />
	<PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.1.38" />
	<PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.1.1" />
	<PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.1.1" />
	<PackageReference Include="System.IO.Pipelines" Version="4.7.0" />
	<PackageReference Include="System.CodeDom" Version="4.7.0" />
	<PackageReference Include="Microsoft.AspNetCore.Connections.Abstractions" Version="2.1.3" />
</ItemGroup>

<ItemGroup Condition="$(TargetFramework) == 'netcoreapp3.1'">
	<FrameworkReference Include="Microsoft.AspNetCore.App" />
	<PackageReference Include="System.CodeDom" Version="4.7.0" />
	<PackageReference Include="Microsoft.AspNetCore.Connections.Abstractions" Version="3.1.32" />
</ItemGroup>

<ItemGroup Condition="$(TargetFramework) == 'net5.0'">
	<FrameworkReference Include="Microsoft.AspNetCore.App" />
	<PackageReference Include="System.CodeDom" Version="5.0.0" />
	<PackageReference Include="Microsoft.AspNetCore.Connections.Abstractions" Version="5.0.17" />
</ItemGroup>
<ItemGroup Condition="$(TargetFramework) == 'net6.0'">
	<FrameworkReference Include="Microsoft.AspNetCore.App" />
	<PackageReference Include="System.CodeDom" Version="6.0.0" />
	<PackageReference Include="Microsoft.AspNetCore.Connections.Abstractions" Version="6.0.27" />
</ItemGroup>
```

andersjonsson: HResult can be used, but that one varies more

akshaybheda: I am making the csproj changes, give me 5  mins will push those

akshaybheda: Done, also incremented the nuget version so you could release

andersjonsson (approved):
Looks good! Thank you!

akshaybheda:
Can you also release this ?

andersjonsson:
> Can you also release this ?

Yup! I'll try to get it released later today

akshaybheda:
Sounds good!! Looking forward to merge and new version

akshaybheda:
Hey, What time do you plan to release it?

PR #1040, opened 2024-03-18
fix(wsdl-asmx): handle nullable enums

Yozer:
When the underlying type for nullable is an enum, we should add nillable attribute


andersjonsson:
Thank you!

PR #1044, opened 2024-03-29
feat(wsdl-meta): allow to customize message names for operation

Yozer:
I have several SOAP services and they generate different message names for operations.
This PR extends SoapCore so users can customize the input/output messages in whatever way they want.

### DIFF
--- a/src/SoapCore.Tests/Wsdl/DefaultWsdlOperationNameGeneratorTests.cs
+++ b/src/SoapCore.Tests/Wsdl/DefaultWsdlOperationNameGeneratorTests.cs
@@ -1,0 +1,40 @@
+using System.ServiceModel;
+using SoapCore.Meta;
+using SoapCore.ServiceModel;
+using SoapCore.Tests.Wsdl.Services;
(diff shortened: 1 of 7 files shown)


andersjonsson:
I'm thinking that this should be used in MetaWCFBodyWriter also. Thoughts?

Yozer:
Yep, definitely someone might want to customize it for WCF. 
I'm not that familiar with the WCFBodyWriter as I only care about the MetaBodyWriter.
I assume we would need a different implementation there as the default names are probably different?

andersjonsson:
As far as I can tell it's the same?
https://github.com/DigDes/SoapCore/blob/dbc85259f4425cbedc7588112107d243cae8cb9d/src/SoapCore/Meta/MetaWCFBodyWriter.cs#L990-L1037

BindingType is confusingly named. It's set to service.GeneralContract.Name in the constructor

Yozer:
Added support for MetaWCFBodyWriter

andersjonsson:
Looks good! Thank you!

PR #1046, opened 2024-03-29
feat(wsdl-meta): add support for XmlArray attribute

Yozer:
XmlElement was supported but array might use XmlArray attribute to specify name.


andersjonsson:
Thanks

PR #1048, opened 2024-03-29
fix:(soap-endpoint): skip duplicated xml parameters

Yozer:
This mimics the behaviour of old System.Web.WebService.
If xml element is duplicated, only first will be used to invoke operation and others will be skipped.


andersjonsson:
Thanks!

PR #1049, opened 2024-03-29
feat(wsdl-meta): support bool-specified pattern

Yozer:
Adds a support for a common pattern in System.Web.WebServices where you could have boolean property that influenced the maxOccurs.
I couldnt find any MS article but this pattern is/as wildly used for XmlSerializer:
https://stackoverflow.com/a/7563231


andersjonsson:
LGTM! Thanks!

PR #1063, opened 2024-05-31
feat(meta-wsdl): add support for soap:header

Yozer:
This implements the soap:header for MetaWritter.
https://learn.microsoft.com/pl-pl/dotnet/api/system.web.services.protocols.soapheader?view=netframework-4.8.1
I created new SoapHeader attribute that mimics the the SoapHeader from net framework.
Users can derive from it and decorate methods on the service.

@andersjonsson what is the recommended way to get the header and inject it into the service?
In my soap service I simply extended the IServiceOperationTuner so it also has ```Message```.
Then I could do that to deserialize my header and inject it:
```csharp
public class SoapHeaderOperationTuner : IServiceOperationTuner
{
    public void Tune(HttpContext httpContext, object serviceInstance, SoapCore.ServiceModel.OperationDescription operation, Message message)
    {
        if (operation.HeaderType is null)
            return;

        var service = (serviceInstance as MyService)!;

        var headerName = operation.HeaderType.Name;
        var contractNamespace = operation.Contract.Namespace;

        var headerIndex = message.Headers.FindHeader(headerName, contractNamespace);
        if (headerIndex < 0) 
            return;

        var header = message.Headers[headerIndex].ToString();
        if (string.IsNullOrEmpty(header))
            return;

        var xml = new XmlDocument();
        xml.LoadXml(header);
        if(xml.DocumentElement == null)
            return;

        using var nodeReader = new XmlNodeReader(xml.DocumentElement);
        var serializer = CachedXmlSerializer.GetXmlSerializer(operation.HeaderType, headerName, contractNamespace);

        var deserializedHeader = serializer.Deserialize(nodeReader);
        if (deserializedHeader is AuthenticationContext authHeader)
            service.AuthenticationContextHeader = authHeader;
    }
}
```

### DIFF
--- a/src/SoapCore.Tests/Wsdl/Services/IServiceWithSoapHeaders.cs
+++ b/src/SoapCore.Tests/Wsdl/Services/IServiceWithSoapHeaders.cs
@@ -1,0 +1,35 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.ServiceModel;
+using System.Text;
+using System.Threading.Tasks;
+using System.Xml;
+using System.Xml.Serialization;
+using SoapCore.ServiceModel;
+#pragma warning disable SA1401 // Fields should be private
+
+namespace SoapCore.Tests.Wsdl.Services;
+[ServiceContract]
+public interface IServiceWithSoapHeaders
+{
+	[OperationContract]
+	[AuthenticationContextSoapHeader]
+	public void Method();
+}
+
+public class ServiceWithSoapHeaders : IServiceWithSoapHeaders
+{
+	public void Method()
+	{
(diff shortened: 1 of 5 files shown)


andersjonsson:
> @andersjonsson what is the recommended way to get the header and inject it into the service? In my soap service I simply extended the IServiceOperationTuner so it also has `Message`. Then I could do that to deserialize my header and inject it:

You could use an ISoapMessageProcessor to parse the header from the Message and then add the header to httpContext.Items

Yozer:
@andersjonsson bump :)

andersjonsson:
Thanks!

PR #1064, opened 2024-05-31
fix: soap action parsing for malformed data

### LINKED ISSUE #1056: CaseInsensitivePath is not respected when comparing the operation name
https://github.com/DigDes/SoapCore/blob/a6474c9e89787477cf5a321e9ccac0a62642b18d/src/SoapCore/SoapEndpointMiddleware.cs#L499

Yozer:
Fixes some malformed requests that clients send and also fixes #1056 as I encountered in on large production service


andersjonsson:
Thanks!

PR #1089, opened 2024-10-07
Remove extra scope while getting IMessageInspector2.

### LINKED ISSUE #1088: Extra scope for IMessageInspector2.
Hi.
I don't understand what the purpose of this change in SoapEndpointMiddleware.cs was and in what cases might an extra scope be needed. But after that there is logical error. If we want to get implementation from DI in service implementation, we will get new implementation of IMessageInspector2 instead of existed. We need it to add some behavior in runtime and accumulate logs (request, some process logs, reply).
I think IMessageInspector2 should be in request scope.

```C#
using (IServiceScope scope = serviceProvider.CreateScope())
{
	var messageInspector2s = scope.ServiceProvider.GetServices<IMessageInspector2>();
	correlationObjects2 = messageInspector2s.Select(mi =>
		(inspector: mi, correlationObject: mi.AfterReceiveRequest(ref requestMessage, _service)))
		.ToList();
}
```

avoronin565:
fix: #1088

### DIFF
--- a/src/SoapCore/SoapEndpointMiddleware.cs
+++ b/src/SoapCore/SoapEndpointMiddleware.cs
@@ -495,16 +495,10 @@ namespace SoapCore
 				throw new ArgumentException($"Unable to handle request without a valid action parameter. Please supply a valid soap action.");
 			}
 
+			var messageInspector2s = serviceProvider.GetServices<IMessageInspector2>();
 			var correlationObjects2 = default(List<(IMessageInspector2 inspector, object correlationObject)>);
 
-			//Not sure that this extra scope is needed...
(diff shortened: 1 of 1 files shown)


andersjonsson:
Yep, I think you're right about the extra scope not being needed

andersjonsson:
Thanks!

avoronin565:
Could you please make release with this fix? Thanks.

avoronin565:
Dear friend, could you please make release please? Need these fix very much. Thank you!

andersjonsson:
> Dear friend, could you please make release please? Need these fix very much. Thank you!

Totally missed this. Sorry.
New release is out now

Issue #1090, opened 2024-10-08
WSDL served from file with 'import' tag can't be parsed because of '&' character


kick2nick:
There is '&' in schemaLocation attribute added during schema loading: https://github.com/DigDes/SoapCore/blob/6d9a2d4a4de34b16273e10112b9d4545fe4709ef/src/SoapCore/Meta/MetaFromFile.cs#L111
Attempts to parse such wsdl are failing. After removing 'import' tags from file wsdl parsed successfully.
Suggestions: 
Replace `&` with `&amp;`
Dont't generate schemaLocation for import tags that already have schema url in namespace attribute.

andersjonsson:
Nice catch! Can you submit a PR?

kick2nick:
Yes, I'll submit 
What do you think about this `Dont't generate schemaLocation for import tags that already have schema url in namespace attribute.` ?

andersjonsson:
> What do you think about this `Dont't generate schemaLocation for import tags that already have schema url in namespace attribute.` ?

I agree with your reasoning. Not sure if it can have other side-effects but it seems unlikely

andersjonsson:
I'll close this since your PR is merged

PR #1091, opened 2024-10-09
Fix wsdl from file generation with `import`/`include` tags

kick2nick:
Issue: https://github.com/DigDes/SoapCore/issues/1090


andersjonsson:
Thanks!

PR #1092, opened 2024-10-15
Fix conflicting variable name in AddSoapMessageProcessor() example

TimSirmovics:
If you copy and paste the `AddSoapMessageProcessor()` example, the declaration of a `message` variable within the method conflicts with the `message` parameter. I have renamed the second instance to `originalMessage`.


andersjonsson:
Nice! Thanks!

PR #1101, opened 2024-12-06
Allows for large files to be processed

EricSteuart:
Checks if the body of the request is a `FileBufferingReadStream` and, if so, uses that instead of using a `MemoryStream`. A `FileBufferingReadStream` is used when `Request.EnableBuffering()` is called. It caches to disk if the stream is large. By using this instead of a `MemoryStream`, an `OutOfMemoryException` can be avoided when processing a very large request.


andersjonsson:
Looks good! Thanks!

PR #1107, opened 2024-12-30
Added support for class inheritance

mdbree-keesing:
**CONTEXT**
I use a property in my model that is based on a base-class. I want the consumer of the SVC to provide (out of the inherited classes) which one it likes to use as an value. This isn't possible since the derived classes aren't included in the SVC output which results in a missing class when generating the model via the Microsoft Visual Studio WCF code generator.

**PROBLEM**
Currently it is checking for the custom attributes of the 'serviceType' but it doesn't check for the attributes on all related types within the same namespace, hence the inherited classes aren't included.

**FIXED**
These changes will include all the related types within the same namespace to resolve this issue. It will allow to use inheritance within the model.


andersjonsson:
Is there a risk of this code being too greedy since it includes all types in the same namespace?

mdbree-keesing:
No I do not believe so, it will not take all the types, only the types from within the namespace of the service type. Also it has to have the attribute 'ServiceKnownType' to be included as extra - if serializing is intended.

andersjonsson:
Fair enough! Thanks!

PR #1110, opened 2025-01-14
Fix SOAP 1.1 fault serialization

astrada:
The SOAP Fault element (in SOAP 1.1) requires that `faultactor` is inserted after `faultstring`, otherwise the SOAP message is invalid (https://schemas.xmlsoap.org/soap/envelope/).

This patch:
- Moves `faultactor` subelement after `faultstring`
- Adds a test to verify that the `Actor` property of the `MessageFault` contained in the catched `FaultException` is correctly populated.


andersjonsson:
Looks good. Thanks!

PR #1112, opened 2025-01-20
Use Path.Combine instead of string concatenation

vincent1405:
Use:
```csharp
Path.Combine(path, meta.WSDLFolder, wsdlfile)
```
instead of
```csharp
path + Path.AltDirectorySeparatorChar + meta.WSDLFolder + Path.AltDirectorySeparatorChar + wsdlfile
```


andersjonsson:
Thanks for the PR. Unfortunately, it doesn't build. Can you fix the errors please?

src/SoapCore/SoapEndpointMiddleware.cs
```diff
@@ -1124,14 +1124,14 @@ private async Task ProcessMetaFromFile(HttpContext httpContext, bool showDocumen
...
-			meta.WSDLFolder = mapping.WSDLFolder;
+			meta.WSDLFolder = mapping.WSDLFolder ?? "";
 			meta.XsdFolder = mapping.SchemaFolder;
 			meta.ServerUrl = GetServerUrl(options, httpContext);
 
 			string wsdlfile = mapping.WsdlFile;
 
-			string path = options.AppPath;
-			string wsdl = await meta.ReadLocalFileAsync(path + Path.AltDirectorySeparatorChar + meta.WSDLFolder + Path.AltDirectorySeparatorChar + wsdlfile);
+			string path = options.AppPath ? "";
```
andersjonsson:
```suggestion
			string path = options.AppPath ?? "";
```

zgabi:
This PR is wrong, please do not merge it.
WSDLFolder can be null, it should not be set to empty string.
Path.Combine does not work with nulls. Probably this is why he set the WSDLFolder to empty string... (And caused a lot of test fails)
The string concatenations works with nulls... there will be double path separators, which is not so nice, but it works with the file operations. (File.ReadAllText(@"C:\Git\1.txt") and File.ReadAllText(@"C:\Git\\\\1.txt") both works)

+1: There are 3 path string concatenations in this file, why just this line was "fixed"? :)

Maybe you can write a helper method which accepts null. But anyway I think this PR makes no sense.

PR #1113, opened 2025-01-24
Call BeforeSendReply even when the service throws an exception

astrada:
Add an exception handler in ProcessMessage to call BeforeSendReply even if the service throws an exception.


andersjonsson:
Thanks!

PR #1114, opened 2025-01-24
HTTP scheme override in the generated WSDL

astrada:
Add an option to override the HTTP scheme in the `soap:address` element of the generated WSDL.


andersjonsson:
LGTM. Thanks!

PR #1118, opened 2025-03-04
Performance work

andersjonsson:


andersjonsson:
My machine is more generous than github actions. I'll have to figure out how to handle multiple implementations without going nuts 😉

PR #1119, opened 2025-03-04
Implement custom message class

andersjonsson:
Proof-of-concept-branch with additional performance work

Right now this is only used for reading the Request message. If it is going to be used for the Response message I have to figure out how this relates to CustomMessage. Should I merge the two implementations? Probably...

This branch uses a custom implementation of Message to make working with it more efficient. I also had to implement a custom XmlReader since the one returned by `XDocument.GetReader()` doesn't support binary data.

All tests pass, but nasty edge case bugs are still possible.

**develop**
| Method    | LoopNum | Mean     | Error     | StdDev    | Gen0      | Allocated |
|---------- |-------- |---------:|----------:|----------:|----------:|----------:|
| EmptyTask | 100     | 1.456 ms | 0.0474 ms | 0.0123 ms |  261.7188 |   1.05 MB |
| Echo      | 100     | 8.745 ms | 0.6351 ms | 0.1649 ms | 1656.2500 |   6.64 MB |

**This branch**
| Method    | LoopNum | Mean     | Error     | StdDev    | Gen0      | Gen1    | Allocated |
|---------- |-------- |---------:|----------:|----------:|----------:|--------:|----------:|
| EmptyTask | 100     | 1.550 ms | 0.0965 ms | 0.1112 ms |  263.6719 |  1.9531 |   1.05 MB |
| Echo      | 100     | 7.110 ms | 0.0646 ms | 0.0718 ms | 1187.5000 | 31.2500 |   4.82 MB |

### DIFF
--- a/src/SoapCore/Extensibility/AuthorizeOperationMessageProcessor.cs
+++ b/src/SoapCore/Extensibility/AuthorizeOperationMessageProcessor.cs
@@ -43,16 +43,16 @@ namespace SoapCore.Extensibility
 		public async Task<Message> ProcessMessage(Message requestMessage, HttpContext httpContext, Func<Message, Task<Message>> next)
 		{
 			var soapAction = HeadersHelper.GetSoapAction(httpContext, ref requestMessage);
-			requestMessage.Headers.Action = soapAction;
+			requestMessage.Headers.Action = soapAction.ToString();
 
-			if (string.IsNullOrEmpty(soapAction))
+			if (soapAction.IsEmpty)
 			{
 				throw new ArgumentException("Unable to handle request without a valid action parameter. Please supply a valid soap action.");
 			}
 
 			if (!_pathTypes.TryGetValue(httpContext.Request.Path.Value.ToLowerInvariant(), out Type serviceType) || !TryGetOperation(soapAction, serviceType, _generateSoapActionWithoutContractName, out var operation))
 			{
-				throw new InvalidOperationException($"No operation found for specified action: {soapAction}");
+				throw new InvalidOperationException($"No operation found for specified action: {soapAction.ToString()}");
 			}
 
 			MethodInfo methodInfo = operation!.DispatchMethod;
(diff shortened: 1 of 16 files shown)


andersjonsson:
@akshaybheda I know you've asked about performance before. Do you have a good scenario for doing some benchmarks on this? Seems to me that the perf is better but I'd like to be sure that it also applies to a real scenario.

akshaybheda:
Let me find some time and do some testing around this

akshaybheda:
This branch has slightly better performance than the latest version 1.2.0
Latest version

image

Your branch with changes
image

src/SoapCore/SoapEndpointMiddleware.cs
```diff
@@ -449,7 +449,7 @@ private async Task<Message> ProcessMessage(Message requestMessage, SoapMessageEn
 		{
 			Message responseMessage;
 			var soapAction = HeadersHelper.GetSoapAction(httpContext, ref requestMessage);
-			requestMessage.Headers.Action = soapAction;
+			requestMessage.Headers.Action = soapAction.ToString();
```
akshaybheda: In the GetSoapAction method, you are already doing the ToString()

src/SoapCore/XDocumentXmlReader.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Linq;
+using System.Text;
+using System.Xml;
+using System.Xml.Linq;
+
+namespace SoapCore
+{
+	public class XDocumentXmlReader : XmlReader
+	{
+		private readonly XDocument _document;
```
akshaybheda: never used

src/SoapCore/XDocumentXmlReader.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Xml;
+using System.Xml.Linq;
+
+namespace SoapCore
+{
+	public class XDocumentXmlReader : XmlReader
```
akshaybheda, line 11: you can also implement IDisposable to dispose the streams

src/SoapCore/Extensibility/AuthorizeOperationMessageProcessor.cs
```diff
@@ -43,7 +43,7 @@ public AuthorizeOperationMessageProcessor(Dictionary<string, Type> pathAndTypes,
 		public async Task<Message> ProcessMessage(Message requestMessage, HttpContext httpContext, Func<Message, Task<Message>> next)
 		{
 			var soapAction = HeadersHelper.GetSoapAction(httpContext, ref requestMessage);
-			requestMessage.Headers.Action = soapAction;
+			requestMessage.Headers.Action = soapAction.ToString();
```
akshaybheda: I don't think this is needed

src/SoapCore/MessageEncoder/SoapMessageEncoder.cs
```diff
@@ -136,19 +136,7 @@ public async Task<Message> ReadMessageAsync(Stream stream, int maxSizeOfHeaders,
...
-			}
-			else
-			{
-				ms = new MemoryStream();
-				await stream.CopyToAsync(ms);
-				ms.Seek(0, SeekOrigin.Begin);
-			}
-
-			XmlReader reader;
+			Message message;
```
akshaybheda, line 139: I don't think you need async Task<Message>, there are no awaits here, you could just return Message and update the callers

src/SoapCore/MessageEncoder/SoapMessageEncoder.cs
```diff
@@ -162,17 +150,23 @@ public async Task<Message> ReadMessageAsync(Stream stream, int maxSizeOfHeaders,
 
 			if (supportXmlDictionaryReader)
 			{
-				reader = XmlDictionaryReader.CreateTextReader(ms, readEncoding, ReaderQuotas, dictionaryReader => { });
+				using (var xdReader = XmlDictionaryReader.CreateTextReader(stream, readEncoding, ReaderQuotas, dictionaryReader => { }))
```
akshaybheda, line 153: nitpick: you can simplify the using

akshaybheda (changes requested):
Did look at some of the changes and added some comments

andersjonsson:
Thank you for taking the time.
I was hoping for a more clear performance win, but I'll take it :)

I also managed to get ParsedMessage working for the response, which resulted in higher memory usage and no speed improvement, so I'll give up on that idea.

I also tried rewriting the DefaultOperationInvoker to use expression trees, but that didn't result in a noticeable win. Maybe with more complex operations... I put that implementation in a gist, in case you want to try it out for your scenario
https://gist.github.com/andersjonsson/004925c250695a44ab7c2f0637dfd465

Rewriting the invocation code to use source generators would most likely be a big benefit, but that would be a big project and would not work for netstandard 2.0.

src/SoapCore/XDocumentXmlReader.cs
andersjonsson: XmlReader already implements IDisposable but I added an overload to dispose the stream

src/SoapCore/MessageEncoder/SoapMessageEncoder.cs
andersjonsson: Fixed

src/SoapCore/SoapEndpointMiddleware.cs
andersjonsson: Good catch! Fixed

akshaybheda (approved):
Look good to me

PR #1137, opened 2025-04-10
added cancellation to readmessage

andersjonsson:


src/SoapCore/ParsedMessage.cs
```diff
@@ -50,9 +47,9 @@ public static async Task<ParsedMessage> FromStreamReaderAsync(StreamReader strea
 			}
 
 #if NETCOREAPP3_0_OR_GREATER
-			var envelope = await XDocument.LoadAsync(stream, LoadOptions.None, CancellationToken.None);
+			var envelope = await XDocument.LoadAsync(stream, LoadOptions.None, ct);
```
akshaybheda: This works only if the user cancel's the request, if user doesn't cancel the request or flood the system with invalid content length headers, this keeps on waiting until IIS or the server doesn't timeout the request. Initially i tried the same code but then ran into same issue

akshaybheda: Also this issue occurs when you give Content-Length that is greater than actual content length. then it waits up

src/SoapCore/MessageEncoder/SoapMessageEncoder.cs
```diff
@@ -128,18 +129,18 @@ public bool IsContentTypeSupported(string contentType, bool checkCharset)
 			return false;
 		}
 
-		public async Task<Message> ReadMessageAsync(PipeReader pipeReader, int maxSizeOfHeaders, string contentType)
+		public async Task<Message> ReadMessageAsync(PipeReader pipeReader, int maxSizeOfHeaders, string contentType, CancellationToken ct)
```
akshaybheda: one other thing i noticed is, we are not utilizing the pipereader, we are anyways converting it back to stream , so we could just send body from the middleware instead of bodyreader

akshaybheda (commented):
I think this works for now, but you still need to add Exception handling since now it throws OperationCancelledException

```
			else if (exception is OperationCanceledException)
			{
				#if NET8_0_OR_GREATER
				status = StatusCodes.Status499ClientClosedRequest;
				#else
				status = StatusCodes.Status408RequestTimeout;
				#endif
			}
```
So maybe add this in the ```CreateErrorResponseMessage``` method and this should all work combined with Net 8 request timeout

src/SoapCore/ParsedMessage.cs
```diff
@@ -49,21 +48,47 @@ public static async Task<ParsedMessage> FromStreamReaderAsync(StreamReader strea
...
 			}
 
+			var pipe = new Pipe();
+
+			var reader = new StreamReader(pipe.Reader.AsStream(), readEncoding);
+
 #if NETCOREAPP3_0_OR_GREATER
-			var envelope = await XDocument.LoadAsync(stream, LoadOptions.None, CancellationToken.None);
+			var envelopeTask = XDocument.LoadAsync(reader, LoadOptions.None, ct);
+			await stream.CopyToAsync(pipe.Writer.AsStream(), ct);
```
akshaybheda, line 58: I think this line should be above the envelopeTask because otherwise it still goes into the same situation

src/SoapCore/ParsedMessage.cs
andersjonsson: Yes, I think that timeouts should be handled with the asp.net core timeout middleware

src/SoapCore/MessageEncoder/SoapMessageEncoder.cs
andersjonsson: Good catch! Fixed

src/SoapCore/ParsedMessage.cs
akshaybheda: @andersjonsson  Did you look at this? Rest everything looks good to me

akshaybheda: NVM i see its not awaited, it works great now

akshaybheda (approved):
Looks good. thanks for the fix

PR #1140, opened 2025-05-12
Fixed matching URL path to service type in Extensibility.AuthorizeOperationMessageProcessor

SerbanApostol:
Previously, the caller of the constructor had the responsibility to make the path for the service to be lowercase. I modified the constructor to make the path lowercase as it is not obvious to the caller that this is necessary.

### DIFF
--- a/src/SoapCore.Tests/Startup.cs
+++ b/src/SoapCore.Tests/Startup.cs
@@ -45,16 +45,17 @@ namespace SoapCore.Tests
 					};
 				});
 
-			services.AddSoapMessageProcessor(new AuthorizeOperationMessageProcessor(new Dictionary<string, Type>
-			{
-				{ "/Service.svc", typeof(TestService) },
-				{ "/ServiceCI.svc", typeof(TestService) },
-				{ "/Service.asmx", typeof(TestService) },
-				{ "/WSA10Service.svc", typeof(TestService) },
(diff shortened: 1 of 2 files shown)


andersjonsson:
There is an option to make path insensitive (SoapCoreOption.CaseInsensitivePath). Would it be a good idea to use that to decide if comparison should be case-insensitive?

Dictionary accepts a comparer in the constructor `new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase)` that makes it easy to vary behavior depending on the option

SerbanApostol:
Thank you for pointing me to the option of supplying a comparer to Dictionary<,>. I removed the foreach that made each key lowercase and instead used the comparer.

andersjonsson:
Sorry for not getting back to you. I haven't had time to work on open source for a while.

I think that you should check the setting SoapCoreOption.CaseInsensitivePath and use that to decide whether to use `StringComparer.Ordinal` or `StringComparer.OrdinalIgnoreCase`

Other than that I think it looks good!

SerbanApostol:
I now notice you mentioned deciding whether to do case insensitive comparison based on the value of SoapCoreOption.CaseInsensitivePath in the first comment as well, sorry about that. I added a constructor that takes an options configuring Action, what I probably should have done the first time around.
I kept the old constructor as well for backwards compatibility, in case anyone else used this in the first place.

andersjonsson:
Thanks!

PR #1141, opened 2025-05-13
Fix handling of trailing slash in remainingPath

danghyan:
Fixed handling of a trailing slash ("/") in remainingPath.
Previously, when a trailing slash was present, an incorrect error related to the Content-Type header was returned.

In my case, all requests ending with a trailing slash resulted in the following error:
`System.ServiceModel.ProtocolException: An HTTP Content-Type header is required for SOAP messaging and none was found.`


andersjonsson:
Thanks!

PR #1145, opened 2025-07-07
Add namespace override support to soap 1.1 faults

### LINKED ISSUE #1143: Setting namespace prefix on SoapFault faultcode
I'm trying to steer namespace prefixes on my responses as I'm re-implementing an existing service. 

I have succeeded to get it to work on my regular responses by overriding the  namespace manager:

``` 
        var soapNamespaceManager = new XmlNamespaceManager(new NameTable());
        soapNamespaceManager.AddNamespace("soap", "http://schemas.xmlsoap.org/soap/envelope/");

        app.UseSoapEndpoint<ISoapEndpoint>(options =>
        {
            options.SoapSerializer = SoapSerializer.XmlSerializer;
            options.XmlNamespacePrefixOverrides = soapNamespaceManager;
        });

```

But I cannot get it to be correct for faults, where the namespace on the faultcode is `s:Server`, but should be `soap:Server`.

```
<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"
               xmlns:xsd="http://www.w3.org/2001/XMLSchema"
               xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <soap:Body>
        <soap:Fault>
            <faultcode>s:Server</faultcode>
            <faultstring>Delivery error</faultstring>
            <detail>
                <fault xmlns="..." xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
                    <code>22</code>
                    <reason>Delivery error</reason>
                </fault>
            </detail>
        </soap:Fault>
    </soap:Body>
</soap:Envelope>
```

If I add `new FaultCode("Server", "http://schemas.xmlsoap.org/soap/envelope/"))` to my FaultException I get `<a:faultcode xmlns:a="http://schemas.xmlsoap.org/soap/envelope/">a:Server</a:faultcode>` which is also off.

It looks like the underlying issue is that the `s:` is hardcoded in https://github.com/DigDes/SoapCore/blob/aa5d590b3542cdf04b58358981cb9381bf699395/src/SoapCore/FaultBodyWriter.cs#L122

Any suggestions on how to oversteer it?

For regular responses it looks like it is the lookup of the prefix that makes it tick: https://github.com/DigDes/SoapCore/blob/aa5d590b3542cdf04b58358981cb9381bf699395/src/SoapCore/CustomMessage.cs#L46

---

An alternative would be for me to be able to re-add `s` as a prefix in `AdditionalEnvelopeXmlnsAttributes`. But those attributes are never passed through to the CustomMessage written in `DefaultFaultExceptionTransformer`. I tried overriding that implementation, but the initializers for the NamespaceManager and the AdditionalEnvelopeXmlnsAttributes are internal on the CustomMessage

https://github.com/DigDes/SoapCore/blob/aa5d590b3542cdf04b58358981cb9381bf699395/src/SoapCore/CustomMessage.cs#L21

staal:
Add support for overriding the default namespaces for soap 1.1 faults. The default is soap envelope namespace `s`, but this allows to override to anything. 

The issue was that the fault code value must also have the same prefix as the soap envelope, but before this was hardcoded to `s:`. 

Resolves https://github.com/DigDes/SoapCore/issues/1143

### DIFF
--- a/src/SoapCore.Tests/RawRequestSoap11Tests.cs
+++ b/src/SoapCore.Tests/RawRequestSoap11Tests.cs
@@ -1,3 +1,4 @@
+using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Text;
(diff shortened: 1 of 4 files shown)


andersjonsson:
There is a conflict in src/SoapCore.Tests/Startup.cs
Can you fix that? I'll get it merged after that

staal:
It is now up to date with develop

andersjonsson:
Thanks!

PR #1146, opened 2025-07-18
Fix: "A concurrent update was performed on this collection and corrupted its state."

JonRobidus:
Fix: "A concurrent update was performed on this collection and corrupted its state."

Don't share XmlNamespaceManager kept in a singleton with request scoped objects that modify it to avoid state corruption.

XmlNamespaceManager Is not thread safe!

### DIFF
--- a/src/SoapCore.Tests/FaultExceptionTransformer/TestFaultExceptionTransformer.cs
+++ b/src/SoapCore.Tests/FaultExceptionTransformer/TestFaultExceptionTransformer.cs
@@ -9,7 +9,7 @@ namespace SoapCore.Tests.FaultExceptionTransformer
 {
 	public class TestFaultExceptionTransformer : IFaultExceptionTransformer
 	{
-		public Message ProvideFault(Exception exception, MessageVersion messageVersion, Message requestMessage, XmlNamespaceManager xmlNamespaceManager)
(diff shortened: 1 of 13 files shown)


nevsnirG:
Perhaps a cleaner fix would be to make the middleware factory based instead of singleton. We had the same bug and wrote our own middleware wrapper which would instantiate and invoke an instance of the SoapCoreMiddleware per request. If the middleware would just be registered as transient then this issue would also be resolved.

janv97:
I get the same error message. Did you guys already implemented a work around or are you still waiting on the merge request to be merged?

JonRobidus:
We implemented a workaround. See issue #743.
This PR is still waiting for approval.

andersjonsson:
Thank you for this PR! Looks good!

This bug was most likely introduced by the caching that I added in https://github.com/DigDes/SoapCore/pull/1123 so the cleanest fix would be to simply remove that, but I prefer this PR since we get to keep the perf benefits of caching.

It would be fairly easy to implement a custom XmlNamespaceManager that uses a ConcurrentDictionary instead of a Dictionary, but that could become a mess if Microsoft ever changes the implementation of XmlNamespaceManager.

So I'll get this PR merged now. Thanks again!

PR #1148, opened 2025-08-04
upgraded to support .net 9

yadavg-2:


mhoerandner:
Will there be an upgrade to .NET 10, and if so, when?

krstns:
> Will there be an upgrade to .NET 10, and if so, when?

have you already used the .NET 9 version?

PR #1155, opened 2025-09-26
fixed generic type serialization in GetSerializedTypeName

xeeqqw:
My contracts contain generic classes.  

__SoapCore__ lib generates the next WSDL schema:  
```xml
<xsd:complexType name="SomeParentObj">
  <xsd:complexContent mixed="false">
    <xsd:extension base="tns:GenericObj`1"> <--- Here is the problem
      <xsd:sequence>
        <xsd:element minOccurs="1" maxOccurs="1" name="Field1" type="xsd:boolean"/>
        <xsd:element minOccurs="1" maxOccurs="1" name="Field2" type="xsd:boolean"/>
        <xsd:element minOccurs="1" maxOccurs="1" name="Field3" type="xsd:boolean"/>
...
```

Attempting to generate `.cs` file using __wsdl.exe__ utility results in an error:  
> Error: Schema item 'complexType' named 'SomeParentObj' from namespace {namespace}
>  - Invalid 'base' attribute: 'The '`' character, hexadecimal value 0x60, cannot be included in a name.'.
>  - The '`' character, hexadecimal value 0x60, cannot be included in a name.

### DIFF
--- a/src/SoapCore.Tests/Wsdl/Services/ComplexTypeWithGeneric.cs
+++ b/src/SoapCore.Tests/Wsdl/Services/ComplexTypeWithGeneric.cs
@@ -1,0 +1,26 @@
+﻿namespace SoapCore.Tests.Wsdl.Services;
+
+public class ComplexTypeWithGeneric
+{
+	public int Id { get; set; }
+
+	public InnerGenericTypeWithSingleArgument<InnerType> SingleGenericArgument { get; set; }
+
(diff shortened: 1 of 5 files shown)


andreycha:
Hi @andersjonsson ,
We'd be happy to get a new version with this fix to stop using a forked version in our product. Thank you!

andersjonsson:
New version will soon be available on nuget

PR #1157, opened 2025-10-10
Enable CA1865-CA1867: Use 'string.Method(char)' instead of 'string.Method(string)' for string with single char

### LINKED ISSUE #1156: Use char instead of string
Due to .NET globalization and ICU please use char instead of string to find for /.

https://github.com/DigDes/SoapCore/blob/59b025b505dc644f9bcfdbc2afde4e88c152b6d9/src/SoapCore/ServiceModel/OperationDescription.cs#L192

## Reference
https://github.com/dotnet/runtime/issues/120520#issuecomment-3379119763

xtqqczze:
Fix https://github.com/DigDes/SoapCore/issues/1156

https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1865-ca1867

### DIFF
--- a/.globalconfig
+++ b/.globalconfig
@@ -1,0 +1,13 @@
+is_global = true
+
+# CA1865: Use char overload
+# https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1865
+dotnet_diagnostic.CA1865.severity = warning
+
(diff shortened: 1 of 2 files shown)


andersjonsson:
Thanks!

PR #1158, opened 2025-10-17
Fix CA1304: Specify CultureInfo

xtqqczze:
https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1304

Follow-up to https://github.com/DigDes/SoapCore/pull/1157

### DIFF
--- a/.globalconfig
+++ b/.globalconfig
@@ -1,5 +1,9 @@
 is_global = true
 
+# CA1304: Specify CultureInfo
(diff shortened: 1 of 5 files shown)


xtqqczze:
@andersjonsson It seems your `.editorconfig` settings are causing a line ending reformat:

https://github.com/DigDes/SoapCore/blob/85169117fb27c5e9b5408257d0a632ba429b6cb7/src/.editorconfig#L3

You should probably configure Git to handle line endings automatically instead.

andersjonsson:
thanks!

Issue #1163, opened 2025-11-12
Fix check warnings in CI


xtqqczze:
- [x] #1172
- [x] [SA1127: Generic type constraints should be on their own line](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1127.md)
- [x] [SA1208: Using directive ordering](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1208.md)
- [ ] [SA1505: An opening brace should not be followed by a blank line](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1505.md)
- [x] #1173
- [ ] #1174
- [ ] #1175

PR #1164, opened 2025-11-12
Fix SA1127: Generic type constraints should be on their own line

xtqqczze:
Contributes to https://github.com/DigDes/SoapCore/issues/1163

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1127.md


andersjonsson:
Thanks!

PR #1167, opened 2025-12-10
Bug fix/race condition in get soap action

### LINKED ISSUE #1166: thread-safety issue in HeadersHelper.GetSoapAction due to double-returning of ArrayPool buffer
A critical thread-safety issue was identified in the HeadersHelper.GetSoapAction method within HeadersHelper.cs. 

The code responsible for parsing the Content-Type header for the soapAction on .NET 8 and later incorrectly returns a shared buffer from ArrayPool<Range>.Shared multiple times. 

This can lead to race conditions and data corruption in multi-threaded environments, as the same buffer might be rented and used by different threads simultaneously.

erossignon:
fixes #1166   

  This commit addresses a critical thread-safety issue in the `GetSoapAction`
    method when parsing the `Content-Type` header on .NET 8 and later.

    The previous implementation could lead to a race condition where a buffer rented
    from `ArrayPool<Range>.Shared` was returned more than once.
    This double-return could cause memory corruption and unpredictable behavior under
    high concurrent loads.

    The fix consists of the following changes:

    - A `try-finally` block is now used to ensure the rented buffer is returned to the
      pool exactly once, regardless of the execution path.
    - The loops for parsing the header now `break` as soon as the `soapAction`
      is found, improving efficiency.
    - Added boundary checks to prevent potential `IndexOutOfRangeException`
      when processing malformed headers.

    Additionally, a new concurrency stress test, `GetSoapAction_IsThreadSafe`, has been added
    to `SoapCore.Tests`.

    This test uses `Parallel.For` to execute a high volume of concurrent calls to `GetSoapAction`,
    verifying that the fix is robust and effectively prevents the race condition.

### DIFF
--- a/src/SoapCore.Tests/HeadersHelperThreadingTest.cs
+++ b/src/SoapCore.Tests/HeadersHelperThreadingTest.cs
@@ -1,0 +1,52 @@
+using System;
+using System.ServiceModel.Channels;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace SoapCore.Tests
+{
+	[TestClass]
+	public class HeadersHelperThreadingTest
+	{
+		[TestMethod]
+		public void GetSoapAction_IsThreadSafe()
+		{
+			// Arrange
+			const int totalOperations = 200_000;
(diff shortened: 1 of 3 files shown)


andersjonsson:
Nice, thank you!

PR #1171, opened 2025-12-15
Fix WSDL schema generator with non-annotated System.Object

maxineal:
XmlSerializer adds empty complex type for System.Object.
It seems like this:
```
<xsd:complexType name="SoapEntityField">
  <xsd:sequence>
    <xsd:element minOccurs="0" maxOccurs="1" name="FieldValue" type="tns:Object"/>
  </xsd:sequence>
</xsd:complexType>
<xsd:complexType name="Object"/>
```

This patch avoid creation empty complex type.
I not work with classic WCF, but i think classic WCF not adds complex type for System.Object, maybe this patch can add backcompability.


andersjonsson:
Thanks!

PR #1177, opened 2026-01-17
Fix XSD schemaLocation when UrlOverride is specified

zgabi:
sWe have a webservice in the following address: https://pannonris.hu/web/PannonRisIf3Ws?import&name=ServiceDefinitions.xml
The application server hosts the service in the /DataExcahngeService location, but in the public service it is overwritten to web/PannonRisIf3Ws

In the url above the schemaLocation is wrong since the overwrite is not applied sometimes:
image

Another small issue is also fixed... with newer .NET (10) I get a warning, that in the HeaderHelper comparing soapAction to null is redundant, since it is a Span, it wont be null.

### DIFF
--- a/src/SoapCore.Tests/WsdlFromFile/WsdlIncludeTests.cs
+++ b/src/SoapCore.Tests/WsdlFromFile/WsdlIncludeTests.cs
@@ -110,7 +110,7 @@ namespace SoapCore.Tests.WsdlFromFile
 			var addresses = _host.ServerFeatures.Get<IServerAddressesFeature>();
 			var address = addresses.Addresses.Single();
 
-			string url = address + "/Service.asmx?xsd&name=echoIncluded.xsd";
+			string url = address + "/Management/Service.asmx?xsd&name=echoIncluded.xsd";
 
 			Assert.IsNotNull(element);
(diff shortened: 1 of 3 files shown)


zgabi:
Test fixed.

I think the test was wrong, since in another test you can see that the service is located at /Management/Service.asmx
https://github.com/DigDes/SoapCore/blob/04b30d49d76da2cbcf3fc09cad23cdd2efa23cbf/src/SoapCore.Tests/WsdlFromFile/WsdlIncludeTests.cs#L51-L74

But in the CheckXSDIncludeXSD test the Management folder was missing:
https://github.com/DigDes/SoapCore/blob/04b30d49d76da2cbcf3fc09cad23cdd2efa23cbf/src/SoapCore.Tests/WsdlFromFile/WsdlIncludeTests.cs#L113

zgabi:
@andersjonsson Could you please check this PR and merge if it seems to be OK?

zgabi:
@andersjonsson Any news about this issue? Should I change anything?

andersjonsson:
Sorry, I haven't been able to spend time on SoapCore for quite a while.
Looks good, thank you!

PR #1178, opened 2026-01-17
Fix support for MessageVersion.Soap12

pcf0:
`MessageVersion.Soap12` was introduced (or to be more precise, the missing API has been added) with `System.ServiceModel.Primitives` version 6.1.0: https://github.com/dotnet/wcf/commit/7b01263abc18ab285921058397525264d94e5e53

For this reason, the comparison with this SOAP version was probably missing in the code, because this cannot be called with the version of .NET Core 3.1.

I have replaced the comparison to the concrete `MessageVersion` with the `EnvelopeVersion` in the code passages, because this is actually the point of interest.

I had noticed this, because when calling the `...?wsdl` URL the error “Unsupported MessageVersion encountered while writing envelope.” was thrown.

### DIFF
--- a/src/SoapCore.Tests/Wsdl/WsdlTests.cs
+++ b/src/SoapCore.Tests/Wsdl/WsdlTests.cs
@@ -50,6 +50,22 @@ namespace SoapCore.Tests.Wsdl
 		}
 
 		[DataTestMethod]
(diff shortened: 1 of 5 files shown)


andersjonsson:
Thank you!

PR #1182, opened 2026-02-05
ArrayOfGuid and guid types in wsdl

psalmnoelff:
Adjusted ClrTypeResolver to return guid on "Guid" types but preserve "string" type by default 
MetaBodyWriter - use proper XML qualified namespace (avoids using guid for ttp://www.w3.org/2001/XMLSchema namespace)
SoapEndpointMiddleware - make the UseMicrosoftGuid static for other method access


psalmnoelff:
Forgot to add - this is only for those using UseMicrosoftGuid = true

psalmnoelff:
Hi, please consider this change thank you

andersjonsson:
Thank you!

PR #1184, opened 2026-04-13
Fix for issue 1183

xeterixon:
Handling of some rare ill-formated request that used to work in previous versions of SoapCore


andersjonsson:
Thank you!

PR #1186, opened 2026-05-04
Fix: WSDL validation fails - XmlDocument complex type has blank name

MarkJHewitt1975:
Added string.IsNullOrEmpty(name) check to Object type


andersjonsson:
thank you!

PR #1189, opened 2026-05-25
XmlSerializer WSDL: Add check for ShouldSerialize method when set minOccurs attribute.

### LINKED ISSUE #1188: ShouldSerialize methods ignored when generating WSDL (minOccurs always 1)
## Expected behavior
When a class has `ShouldSerialize{PropName}()` returning `false`, the WSDL should generate that element with `minOccurs="0"` (like XmlSerializer does).

## Actual behavior
The element always has `minOccurs="1"`, ignoring the `ShouldSerialize` method.

## Minimal reproduction
```csharp
public class MyRequest
{
    public int? OptionalValue { get; set; }

    public bool ShouldSerializeOptionalValue() => OptionalValue.HasValue;
}
```
Generated WSDL excerpt:
```xml
<xs:element name="OptionalValue" type="xs:int" minOccurs="1" .../>
```

Expected WSDL excerpt:
```xml
<xs:element name="OptionalValue" type="xs:int" minOccurs="0" .../>
```

avoronin565:
When generating WSDL, check `ShouldSerialize{PropName}` methods to determine if a property should be optional.

Previously, all value type properties had `minOccurs="1"`. Now, if `ShouldSerialize{PropName}` exists the element gets `minOccurs="0"`.

Added a new test `CheckShouldSerializeMemberSerialization()` to verify.

Fixes #1188

### DIFF
--- a/src/SoapCore.Tests/Wsdl/Services/IShouldSerializeTypeMemberService.cs
+++ b/src/SoapCore.Tests/Wsdl/Services/IShouldSerializeTypeMemberService.cs
@@ -1,0 +1,20 @@
+using System;
+using System.ServiceModel;
+
+namespace SoapCore.Tests.Wsdl.Services
+{
+	[ServiceContract]
+	public interface IShouldSerializeTypeMemberService
+	{
+		[OperationContract]
+		TypeWithShouldSerializeMember Method();
+	}
+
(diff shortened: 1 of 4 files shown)


andersjonsson:
Thanks!

PR #1193, opened 2026-09-11
Fix duplicate fault response on the error path #1192

### LINKED ISSUE #1192: Fault path calls CreateErrorResponseMessage twice: error logged twice, ProvideFault invoked twice, and the message passed to BeforeSendReply is not the one sent
### Summary

When a service operation throws, the fault response is built twice. The error is written to the log twice, `IFaultExceptionTransformer.ProvideFault` is invoked twice, and the message that `IMessageInspector2.BeforeSendReply` receives is not the message that is sent to the client.

### Code path

`ProcessMessage` catches the exception, builds a fault response, notifies the message inspectors, and rethrows (`src/SoapCore/SoapEndpointMiddleware.cs`, tag `v1.2.1.15`):

```csharp
catch (Exception ex)
{
    responseMessage = CreateErrorResponseMessage(ex, serviceProvider, requestMessage, messageEncoder, httpContext);

    correlationObjects2.ForEach(mi => mi.inspector.BeforeSendReply(ref responseMessage, _service, mi.correlationObject));

    throw;
}
```

Because it rethrows, the exception reaches the `catch` in `ProcessOperation`, which calls `CreateErrorResponseMessage` again:

```csharp
catch (Exception ex)
{
    responseMessage = CreateErrorResponseMessage(ex, serviceProvider, requestMessage, messageEncoder, httpContext);
}
```

That second message is the one written to the response. `CreateErrorResponseMessage` begins with

```csharp
_logger.LogError(exception, "An error occurred processing the message");
```

and calls `faultExceptionTransformer.ProvideFault(...)`.

### Observed

Single POST to an operation that throws `InvalidOperationException("boom")`, SoapCore 1.2.1.15 on net8.0, `SoapSerializer.XmlSerializer`, `MessageVersion.Soap12WSAddressingAugust2004`:

| | Observed |
| --- | --- |
| `Error` entries `"An error occurred processing the message"` from `SoapCore.SoapEndpointMiddleware` | 2 |
| `IFaultExceptionTransformer.ProvideFault` invocations | 2 |
| `IMessageInspector2.AfterReceiveRequest` invocations | 1 |
| `IMessageInspector2.BeforeSendReply` invocations | 1 |
| Header added by `BeforeSendReply` present in the fault received by the client | no |

### Expected

One fault response per faulting request: the error logged once, `ProvideFault` invoked once, and the message handed to `BeforeSendReply` being the message that is sent.

### Affected versions

`ProcessMessage` in `1.1.0.51` has no `catch` around the operation invocation, only `try`/`finally`. The `catch` shown above is present in `1.1.0.53`, `1.2.0.0`, `1.2.1.0` and `1.2.1.15` (each read at its tag). `1.1.0.52` was not checked.

### Reproduce

`Repro.csproj`

```xml
<Project Sdk="Microsoft.NET.Sdk.Web">
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="SoapCore" Version="1.2.1.15" />
  </ItemGroup>
</Project>
```

`Program.cs`

```csharp
using System.ServiceModel;
using System.ServiceModel.Channels;
using SoapCore;
using SoapCore.Extensibility;
using SoapCore.ServiceModel;

[ServiceContract(Namespace = "urn:repro")]
public interface IReproService
{
    [OperationContract]
    string Boom(string input);
}

public class ReproService : IReproService
{
    public string Boom(string input) => throw new InvalidOperationException("boom");
}

public class CountingInspector : IMessageInspector2
{
    public static int AfterReceiveCalls;
    public static int BeforeSendReplyCalls;

    public object AfterReceiveRequest(ref Message message, ServiceDescription serviceDescription)
    {
        Interlocked.Increment(ref AfterReceiveCalls);
        return null;
    }

    public void BeforeSendReply(ref Message reply, ServiceDescription serviceDescription, object correlationState)
    {
        Interlocked.Increment(ref BeforeSendReplyCalls);
        reply.Headers.Add(MessageHeader.CreateHeader("InspectorStamp", "urn:repro", "was-here"));
    }
}

// Only used to count ProvideFault; register before AddSoapCore so the default is not added.
public class CountingFaultTransformer : IFaultExceptionTransformer
{
    public static int ProvideFaultCalls;

    public Message ProvideFault(Exception exception, MessageVersion messageVersion, Message requestMessage, ConcurrentXmlNamespaceLookup xmlNamespaceLookup)
    {
        Interlocked.Increment(ref ProvideFaultCalls);
        return Message.CreateMessage(messageVersion, action: null);
    }
}

public static class Program
{
    public static void Main(string[] args)
    {
        var builder = WebApplication.CreateBuilder(args);
        builder.Logging.ClearProviders();
        builder.Logging.AddSimpleConsole(o => o.SingleLine = true);

        if (args.Contains("--count-transformer"))
        {
            builder.Services.AddSingleton<IFaultExceptionTransformer, CountingFaultTransformer>();
        }

        builder.Services.AddSoapCore();
        builder.Services.AddSingleton<IReproService, ReproService>();
        builder.Services.AddSoapMessageInspector<CountingInspector>();

        var app = builder.Build();
        app.UseRouting();

        app.MapGet("/counters", () =>
            $"AfterReceiveRequest={CountingInspector.AfterReceiveCalls} "
            + $"BeforeSendReply={CountingInspector.BeforeSendReplyCalls} "
            + $"ProvideFault={CountingFaultTransformer.ProvideFaultCalls}");

        ((IApplicationBuilder)app).UseSoapEndpoint<IReproService>(
            path: "/Service.svc",
            new SoapEncoderOptions { MessageVersion = MessageVersion.Soap12WSAddressingAugust2004 },
            SoapSerializer.XmlSerializer);

        app.Run();
    }
}
```

Request:

```
POST /Service.svc
Content-Type: application/soap+xml; charset=utf-8; action="urn:repro/IReproService/Boom"

<s:Envelope xmlns:s="http://www.w3.org/2003/05/soap-envelope"
            xmlns:a="http://schemas.xmlsoap.org/ws/2004/08/addressing">
  <s:Header>
    <a:Action s:mustUnderstand="1">urn:repro/IReproService/Boom</a:Action>
    <a:To s:mustUnderstand="1">http://localhost:5000/Service.svc</a:To>
  </s:Header>
  <s:Body>
    <Boom xmlns="urn:repro"><input>x</input></Boom>
  </s:Body>
</s:Envelope>
```

Run without arguments to see the two log entries, the single `BeforeSendReply` call and the delivered fault without `InspectorStamp`. Run with `--count-transformer` to see `ProvideFault=2`; that run replaces the fault body, so use the two runs separately.
`GET /counters` returns the counters.

juergotter:
Fixes #1192.

### Change

The `catch` in `ProcessMessage` is removed, and `IMessageInspector2.BeforeSendReply` is invoked from the `catch` in `ProcessOperation`, on the message that is written to the response.

The correlation objects are created in `ProcessMessage`, so `ProcessOperation` creates the list and passes it in. `AfterReceiveRequest` stays inside the `ISoapMessageProcessor` pipe, and `ProcessMessage` is private, so no public API is affected. The list is cleared once the inspectors have seen the reply, so a failure later in the response path cannot notify them a second time.

### Measured

Repro from the issue, single POST to an operation that throws:

| | `v1.2.1.15` | this branch |
| --- | --- | --- |
| `Error` entries "An error occurred processing the message" | 2 | 1 |
| `IFaultExceptionTransformer.ProvideFault` invocations | 2 | 1 |
| `IMessageInspector2.AfterReceiveRequest` invocations | 1 | 1 |
| `IMessageInspector2.BeforeSendReply` invocations | 1 | 1 |
| Header added by `BeforeSendReply` present in the fault received by the client | no | yes |

### Adjacent paths

| Path | Behaviour |
| --- | --- |
| exception before any `AfterReceiveRequest` | no inspector notified, unchanged |
| one inspector throws in `AfterReceiveRequest` | the inspectors before it are notified, keeping their calls paired |
| `ISoapMessageProcessor` swallows the exception and returns its own message | no fault built, so no log entry and no notification |
| inspector throws in `BeforeSendReply` | the exception leaves the middleware, so the client gets a plain 500 instead of a fault |

### Tests

`src/SoapCore.Tests/DuplicateFault` runs a `TestServer` with a counting `IFaultExceptionTransformer`, an `IMessageInspector2` that stamps a header onto the reply, and an `ILoggerProvider` that records `Error` entries. One faulting request asserts the rows of the table above. A fifth test asserts that a successful operation followed by a throwing `IAsyncMessageFilter.OnResponseExecuting` notifies the inspectors once.

Three of the five fail on `develop`.

##### Note

`dotnet test src/SoapCore.Tests`: 292 of 293 pass. `TestIssue1183` fails on `develop` as well.

### DIFF
--- a/src/SoapCore.Tests/DuplicateFault/CountingFaultExceptionTransformer.cs
+++ b/src/SoapCore.Tests/DuplicateFault/CountingFaultExceptionTransformer.cs
@@ -1,0 +1,25 @@
+using System;
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using SoapCore.Extensibility;
+
+namespace SoapCore.Tests.DuplicateFault
+{
+	public class CountingFaultExceptionTransformer : IFaultExceptionTransformer
+	{
+		private int _provideFaultCallCount;
+
+		public int ProvideFaultCallCount => _provideFaultCallCount;
+
+		public Message ProvideFault(Exception exception, MessageVersion messageVersion, Message requestMessage, ConcurrentXmlNamespaceLookup xmlNamespaceLookup)
+		{
+			_provideFaultCallCount++;
+
+			var faultException = new FaultException(new FaultReason(exception.Message), new FaultCode("Sender"), null);
+			var messageFault = faultException.CreateMessageFault();
+			var bodyWriter = new MessageFaultBodyWriter(messageFault, messageVersion);
+
+			return Message.CreateMessage(messageVersion, null, bodyWriter);
+		}
+	}
+}
--- a/src/SoapCore.Tests/DuplicateFault/CountingLoggerProvider.cs
+++ b/src/SoapCore.Tests/DuplicateFault/CountingLoggerProvider.cs
@@ -1,0 +1,71 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Logging;
+
+namespace SoapCore.Tests.DuplicateFault
+{
+	public sealed class CountingLoggerProvider : ILoggerProvider
+	{
+		private readonly List<string> _errors = new List<string>();
+
+		public IReadOnlyList<string> Errors
+		{
+			get
+			{
+				lock (_errors)
+				{
+					return _errors.ToList();
+				}
+			}
+		}
+
+		public ILogger CreateLogger(string categoryName)
+		{
+			return new CountingLogger(this);
+		}
+
+		public void Dispose()
+		{
+		}
+
+		private void AddError(string message)
+		{
+			lock (_errors)
+			{
+				_errors.Add(message);
+			}
+		}
+
+		private sealed class CountingLogger : ILogger
+		{
+			private readonly CountingLoggerProvider _provider;
+
+			public CountingLogger(CountingLoggerProvider provider)
+			{
+				_provider = provider;
+			}
+
+			public IDisposable BeginScope<TState>(TState state)
+				where TState : notnull
+			{
+				return null;
+			}
+
+			public bool IsEnabled(LogLevel logLevel)
+			{
+				return true;
+			}
+
+			public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+			{
+				if (logLevel != LogLevel.Error)
+				{
+					return;
+				}
+
+				_provider.AddError(formatter(state, exception));
+			}
+		}
+	}
+}
--- a/src/SoapCore.Tests/DuplicateFault/DuplicateFaultTests.cs
+++ b/src/SoapCore.Tests/DuplicateFault/DuplicateFaultTests.cs
@@ -1,0 +1,132 @@
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace SoapCore.Tests.DuplicateFault
+{
+	// https://github.com/DigDes/SoapCore/issues/1192
+	[TestClass]
+	public class DuplicateFaultTests
+	{
+		private const string FaultMessage = "Test fault message";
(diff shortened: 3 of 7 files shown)


andersjonsson:
Great! Thank you! Will make a new release shortly

